PR #37, opened 2015-06-24
Added time signatures

amirsaffari:
I needed to be able to read and write correct time signatures, so I added the functionality. I tested it briefly and it seems to be fine. 

Thanks, Amir

### DIFF
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -79,6 +79,8 @@ class PrettyMIDI(object):
             # Compute the tick scale for the provided initial tempo
             # and let the tick scale start from 0
             self.__tick_scales = [(0, 60.0/(initial_tempo*self.resolution))]
+            # Set the time signature to 4 / 4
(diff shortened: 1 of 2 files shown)


craffel:
Thanks for this Amir.  My intuition, however, is that we should store time signature events as notes, pitch bends, and control changes are stored - that is, create a new class which holds the information about each time signature change, instead of representing them as an array.  Also, I don't think we need to store the beat and bar duration, as that could be deduced fairly readily from the tempo information, correct?  It would then make sense to me to have a function which calculated this information.  I'm not terribly familiar with the time signature events, so I may be out of my element here, but that's my intuition.

craffel:
Resolved in a53537b5825efd34b8210cb958c919efc166421f.

amirsaffari:
Hi Colin, 

Thanks for incorporating the time signatures and apologies for not being able to get back to you on the previous pull request and subsequent steps, it's been very busy :)

Thanks, Amir

PR #39, opened 2015-08-04
Key signature and Time signature plus other additions

rafaelvalle:
General formatting.
Creating classes and routines related to Key and Time signature changes. They both depend on information provided by _load_tempo_changes and self._update_tick_to_time(max_tick).
Added parameter not to include_pitch_bends on chromagrams and piano rollls.
Added routines for converting between different types of key notation.

### DIFF
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -54,14 +54,16 @@ class PrettyMIDI(object):
         if midi_file is not None:
             # Load in the MIDI data using the midi module
             midi_data = midi.read_midifile(midi_file)
+
             # Convert tick values in midi_data to absolute, a useful thing.
             midi_data.make_ticks_abs()
 
             # Store the resolution for later use
             self.resolution = midi_data.resolution
 
-            # Populate the list of tempo changes (tick scales)
-            self._load_tempo_changes(midi_data)
+            # Populate the list of tempo (tick scales), key and time signatures changes
+            self._load_metadata(midi_data)
+
             # Update the array which maps ticks to time
             max_tick = max([max([e.tick for e in t]) for t in midi_data]) + 1
             # If max_tick is huge, the MIDI file is probably corrupt
@@ -70,6 +72,7 @@ class PrettyMIDI(object):
                 raise ValueError(('MIDI file has a largest tick of {},'
                                   ' it is likely corrupt'.format(max_tick)))
             self._update_tick_to_time(max_tick)
+
             # Check that there are only tempo change events on track 0
             if sum([sum([event.name == 'Set Tempo' for event in track])
                     for track in midi_data[1:]]):
@@ -77,9 +80,22 @@ class PrettyMIDI(object):
                                "  This is not a valid type 0 or type 1 MIDI "
                                "file.  Timing may be wrong."), RuntimeWarning)
 
-            # Populate the list of instruments, key and time signatures
-            self._load_events(midi_data)
+            # Check that there are only key signature change events on track 0
+            if sum([sum([isinstance(event, midi.events.KeySignatureEvent) for event in track])
+                    for track in midi_data[1:]]):
+                warnings.warn(("Key signature change events found on non-zero tracks."
+                               "  This is not a valid type 0 or type 1 MIDI "
+                               "file.  Key Signature may be wrong."), RuntimeWarning)
 
+            # Check that there are only time signature change events on track 0
+            if sum([sum([isinstance(event, midi.events.TimeSignatureEvent) for event in track])
+                    for track in midi_data[1:]]):
+                warnings.warn(("Time signature change events found on non-zero tracks."
+                               "  This is not a valid type 0 or type 1 MIDI "
+                               "file.  Time Signature may be wrong."), RuntimeWarning)
+
+            # Populate the list of instruments
+            self._load_instruments(midi_data)
 
         else:
             self.resolution = resolution
@@ -91,19 +107,21 @@ class PrettyMIDI(object):
             # Empty instruments list
             self.instruments = []
 
-    def _load_events(self, midi_data):
-      """Populates the list of instruments, key signatures and tempo changes.
+    def _load_metadata(self, midi_data):
+        """Populates the lists of tempo, key and time signature changes
 
-      Parameters
-      ----------
-      midi_data : midi.FileReader
-        MIDI object from which data will be read
+        Populates self.__tick_scales with tuples of (tick, tick_scale),
+        Populates self.__time_signatures with TimeSignature objects. It can be accessed by the get_time_signatures method.
+        Populates self.__key_changes with KeySignature objects. It can be accessed by the get_key_signatures method.
 
-      """
-      self._load_instruments(midi_data)
-      self._load_key_changes(midi_data)
-      self._load_time_signatures(midi_data)
+        Parameters
+        ----------
+            - midi_data : midi.FileReader
+                MIDI object from which data will be read
+        """
 
+        #list to store key signature changes
+        self.__key_changes = []
 
     def _load_tempo_changes(self, midi_data):
         """Populates self.__tick_scales with tuples of (tick, tick_scale)
@@ -114,6 +132,10 @@ class PrettyMIDI(object):
             MIDI object from which data will be read
 
         """
+
+        #list to store time signatures changes
+        self.__time_signatures = []
+
         # MIDI data is given in "ticks".
         # We need to convert this to clock seconds.
         # The conversion factor involves the BPM, which may change over time.
@@ -126,7 +148,7 @@ class PrettyMIDI(object):
         # Everyone ignores type 2.
         # So, just look at events on track 0
         for event in midi_data[0]:
-            if event.name == 'Set Tempo':
+            if isinstance(event, midi.events.SetTempoEvent):
                 # Only allow one tempo change event at the beginning
                 if event.tick == 0:
                     bpm = event.get_bpm()
@@ -138,44 +160,15 @@ class PrettyMIDI(object):
                     # Ignore repetition of BPM, which happens often
                     if tick_scale != last_tick_scale:
                         self.__tick_scales.append((event.tick, tick_scale))
-
-    def _load_key_changes(self, midi_data):
-        """Populates self.__key_changes with KeySignature objects
-
-        Data can accessed through the get_key_changes method.
-
-        Parameters
-        ---------
-            midi_data : midi.FileReader
-                MIDI object from which data will be read
-        """
-
-        self.__key_changes = []
-        for event in midi_data[0]:
-            if isinstance(event, midi.events.KeySignatureEvent):
-                key_number = midi_key_to_pretty_key(event)
-                time = self.__tick_to_time[event.tick]
-                key_obj = KeySignature(key_number, time)
+            elif isinstance(event, midi.events.KeySignatureEvent):
+                key_obj = KeySignature(
+                    midi_key_to_key_name(event),
+                    self.__tick_to_time[event.tick])
                 self.__key_changes.append(key_obj)
-
-    def _load_time_signatures(self, midi_data):
-        """Populates self.__time_signatures with TimeSignature objects
-
-        Data can accessed through the get_time_signatures method.
-
-        Parameters
-        ----------
-            midi_data : midi.FileReader
-                MIDI object from which data will be read
-        """
-
-        self.__time_signatures = []
-        for event in midi_data[0]:
-            if isinstance(event, midi.events.TimeSignatureEvent):
-                numerator = event.get_numerator()
-                denominator = event.get_denominator()
-                time = self.__tick_to_time[event.tick]
-                ts_obj = TimeSignature(numerator, denominator, time)
+            elif isinstance(event, midi.events.TimeSignatureEvent):
+                ts_obj = TimeSignature(
+                    event.get_numerator(), event.get_denominator(),
+                    self.__tick_to_time[event.tick])
                 self.__time_signatures.append(ts_obj)
 
     def _update_tick_to_time(self, max_tick):
@@ -596,7 +589,7 @@ class PrettyMIDI(object):
 
 
     def get_key_changes(self):
-        """Returns an array with KeySignature objects acquired directly from the midi file.
+        """Returns a numpy array with KeySignature objects acquired directly from the midi file.
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


pretty_midi/pretty_midi.py
```diff
@@ -71,6 +81,24 @@ def __init__(self, midi_file=None, resolution=220, initial_tempo=120.):
                                "  This is not a valid type 0 or type 1 MIDI "
                                "file.  Timing may be wrong."), RuntimeWarning)
 
+            # Check that there are only key signature change events on track 0
+            if sum([sum([isinstance(event, midi.events.KeySignatureEvent)
```
craffel: Let's combine these three checks into a single one - `isinstance` can take a `tuple` as its second argument, so the call would look like `isinstance(event, (midi.events.SetTempoEvent, midi.events.KeySignatureEvent, midi.events.TimeSignatureEvent)` and you could change the warning to say something like "Key signature, set tempo, or time signature events found on non-zero...".

pretty_midi/pretty_midi.py
```diff
@@ -104,7 +133,7 @@ def _load_tempo_changes(self, midi_data):
         # Everyone ignores type 2.
         # So, just look at events on track 0
         for event in midi_data[0]:
-            if event.name == 'Set Tempo':
+            if isinstance(event, midi.events.SetTempoEvent):
```
craffel: This is better, thanks.

pretty_midi/pretty_midi.py
```diff
@@ -117,6 +146,38 @@ def _load_tempo_changes(self, midi_data):
...
+            It can be accessed by the get_key_signatures method.
+
+        Parameters
+        ----------
+            - midi_data : midi.FileReader
+                MIDI object from which data will be read
+        """
+
+        # list to store key signature changes
+        self.__key_changes = []
```
craffel: I think these can/should actually just be public.  The reason `__tick_scales` is private is that we don't want the user to mess with it, but key changes and time signatures should be freely gettable/settable right?

pretty_midi/pretty_midi.py
```diff
@@ -504,13 +571,38 @@ def get_chroma(self, fs=100, times=None):
...
+            fs=fs,
+            times=times,
+            include_pitch_bends=include_pitch_bends)
         # Fold into one octave
         chroma_matrix = np.zeros((12, piano_roll.shape[1]))
         for note in range(12):
             chroma_matrix[note, :] = np.sum(piano_roll[note::12], axis=0)
         return chroma_matrix
 
+    def get_key_signatures(self):
```
craffel: If __key_changes and __time_signatures are public, we don't need these getter functions.

pretty_midi/pretty_midi.py
```diff
@@ -1258,6 +1353,180 @@ def __repr__(self):
...
+    --------
+    Instantiate a TimeSignature object with 6/8 time signature at 3.14 seconds
+    >>> ts = TimeSignature(6, 8, 3.14)
+    >>> print ts
+    6/8 at 3.14 seconds
+
+    """
+
+    def __init__(self, numerator, denominator, time):
+        assert isinstance(numerator, int), \
```
craffel: We can be more strict here, right?  I.e. the numerator and denominator should greater than 0.

pretty_midi/pretty_midi.py
```diff
@@ -1258,6 +1353,180 @@ def __repr__(self):
...
+        self.key_number = key_number
+        self.time = time
+
+    def __repr__(self):
+        return '%s at %.2f seconds' % (
+            key_number_to_key_string(self.key_number),
+            self.time)
+
+
+def key_number_to_key_string(key_number):
```
craffel: Can we call this `key_number_to_key_name`?

pretty_midi/pretty_midi.py
```diff
@@ -1258,6 +1353,180 @@ def __repr__(self):
...
+        return keys[key_idx] + ' Major'
+    elif mode == 1:
+        # preference to C#, F# and G# minor
+        if key_idx in [1, 6, 8]:
+            return keys[key_idx-1] + '# minor'
+        else:
+            return keys[key_idx] + ' minor'
+
+
+def key_string_to_key_number(key_string):
```
craffel: `key_name_to_key_number`

pretty_midi/pretty_midi.py
```diff
@@ -1623,3 +1892,36 @@ def semitones_to_pitch_bend(semitones, semitone_range=2.):
             MIDI pitch bend amount, in [-8192, 8191]
     '''
     return int(8192*(semitones/semitone_range))
+
+
+def midi_key_to_key_name(key_signature_event):
```
craffel: So, an assumption of `pretty_midi` is that someone using it will never have to use any functionality from the `midi` module.  The conversion functions I made are mostly general-purpose convenience.  Since this is really only used within `_load_metadata`, can we put this in there instead?

craffel:
Thanks for this Rafael.  I made a bunch of line comments, I'll check it out locally and try it a bit.  A high level comment - it looks like you didn't pull the most recent commit (changing docstring formats) before applying these changes, so you'll need to fetch and rebase before I can merge.

pretty_midi/pretty_midi.py
```diff
@@ -456,7 +519,7 @@ def get_onsets(self):
         # Return them sorted (because why not?)
         return np.sort(onsets)
 
-    def get_piano_roll(self, fs=100, times=None):
+    def get_piano_roll(self, fs=100, times=None, include_pitch_bends=True):
```
craffel: Looks like you didn't add `include_pitch_bends` to the docstrings.  You'll need to rebase first before you do that anyways.  Or, we can handle that in a separate pull request, as it's independent of the key/time signature stuff.

pretty_midi/pretty_midi.py
```diff
@@ -869,45 +961,45 @@ def get_piano_roll(self, fs=100, times=None):
...
-            bent_roll = np.zeros(piano_roll[:, bend_range].shape)
-            # Easiest to process differently depending on bend sign
-            if start_bend.pitch >= 0:
-                # First, pitch shift by the int amount
-                if bend_int is not 0:
-                    bent_roll[bend_int:] = piano_roll[:-bend_int, bend_range]
+        if include_pitch_bends:
+            # Process pitch changes
+            # Need to sort the pitch bend list for the following to work
+            ordered_bends = sorted(self.pitch_bends, key=lambda bend: bend.time)
```
craffel: There are a bunch of lines in here which are now > 79 chars - maybe your PEP-8 checker isn't picking up on this.

pretty_midi/pretty_midi.py
```diff
@@ -1258,6 +1353,180 @@ def __repr__(self):
...
+
+    """
+
+    def __init__(self, numerator, denominator, time):
+        assert isinstance(numerator, int), \
+            '%s is not a recognized `numerator`' % str(type(numerator))
+        assert isinstance(denominator, int), \
+            '%s is not a recognized `denominator`' % str(type(denominator))
+        assert isinstance(time, float), \
+            '%s is not a recognized `time`' % str(type(key_number))
```
craffel: `key_number` isn't defined, I think you meant `time`.

pretty_midi/pretty_midi.py
```diff
@@ -1258,6 +1353,180 @@ def __repr__(self):
...
+    assert key_str[0] in note_names_pc, \
+        'Key %s is not recognized' % key_str[0]
+    assert mode_str in modes, \
+        'Mode is not recognized'
+
+    # lookup dictionary
+    key_number = note_names_pc[key_str[0]]
+
+    # offset key_index according to sharp or flat key
+    key_offset = 0
```
craffel: `key_offset` is defined but never used.

pretty_midi/pretty_midi.py
rafaelvalle: I was trying not to touch your code. I'll fix that.

rafaelvalle:
I'm rebasing my code. Do you want me to issue another pull request or something else?

pretty_midi/pretty_midi.py
```diff
@@ -1258,6 +1353,180 @@ def __repr__(self):
...
+            '%s is not a recognized `denominator`' % str(type(denominator))
+        assert isinstance(time, float), \
+            '%s is not a recognized `time`' % str(type(key_number))
+
+        self.numerator = numerator
+        self.denominator = denominator
+        self.time = time
+
+    def __repr__(self):
+        return '%d/%d at %.2f seconds' % (
```
craffel: `__repr__` is meant to produce code which can re-create the object (see `Note`) for example, so this should be changed to something like 
`"TimeSignature(numerator={}, denominator={}, time={})".format(self.numerator, self.denominator, self.time)`

craffel:
> I'm rebasing my code. Do you want me to issue another pull request or something else?

No, should be fine to use this PR, thanks.

pretty_midi/pretty_midi.py
rafaelvalle: I think they should be private.
If the midi file was written properly, that means the user shouldn't have a need to set the time nor key signatures.

If the midi file is a transcription from someone's performance on an instrument, then the user might have the need to change it. In this case, the user can use a setter method, which enforces the correct data structure, data types, etc...

What do you think?

pretty_midi/pretty_midi.py
rafaelvalle: Ops. Fixed

pretty_midi/pretty_midi.py
rafaelvalle: Yes, much better!

pretty_midi/pretty_midi.py
rafaelvalle: OK.

pretty_midi/pretty_midi.py
rafaelvalle: Fixed.

pretty_midi/pretty_midi.py
rafaelvalle: Fixed.

pretty_midi/pretty_midi.py
rafaelvalle: Changing TimeSignature and KeySignature accordingly

pretty_midi/pretty_midi.py
rafaelvalle: Combined.

pretty_midi/pretty_midi.py
craffel: > If the midi file is a transcription from someone's performance on an instrument, then the user might have the need to change it.

Yes, `pretty_midi` is meant to handle the case for MIDI file creation as well as reading/analyzing/modification, so I think we need to provide a way for these lists to be modified.  Currently, users can put whatever they want into, for example, the `.instruments` list (and the `Instrument.notes` list, etc).  We could provide setters which do sanity checks, but the Python way is usually just to provide things publicly for people to do the sane thing with, rather than use private members and setters/getters as in, say, C++.

rafaelvalle: I think that providing setters is a safer way, as it ensures that the methods that rely on time and signature won't fail or raise exceptions. For key and time signature changes, the sanity checks should be straight forward to implement and easier than for Instruments.

craffel: Well, all of them would basically just need to check that the thing being added is an instance of what the list is supposed to hold.  Usually in Python you let the user shoot themselves in the foot as far as properties go, and avoid getter/setters whenever possible; see e.g. http://stackoverflow.com/questions/1554546/when-and-how-to-use-the-builtin-function-property-in-python

If you really believe we need getter/setters with sanity checks, we can do them but only as properties, and we'd need them for all of the list attributes.

rafaelvalle: Let's do as you say, which is to transfer responsibility to the users, let them shoot themselves in the foot and make it public.

craffel: Sounds good to me.

rafaelvalle:
Pushed updated files with the changes we discussed, including pep 8 modifications

craffel:
Hey Rafael, I'd like to make a few minor edits and merge, but I still can't apply your changes without some hefty merge resolution because this PR still has a series of commits which didn't include the numpydoc changes.  Do you want to squash all your WIP commits to a single commit (via a `git rebase -i`) and then `push -f` your changes?  Or, just create a new separate commit/PR with the changes you introduced?

rafaelvalle:
Hey Collin. I've just squashed the WIP commits and pushed.

craffel:
Hm, it's still showing up as 22 commits.  Does git status tell you you have diverged?  If not, the squash was probably not successful.

rafaelvalle:
Can you see these commits?

d4167a9
https://github.com/rafaelvalle/pretty-midi/commit/d4167a9d9634d3552326f03401c32f81b97402b0

or 
8935bad
https://github.com/rafaelvalle/pretty-midi/commit/8935bada20aa4fc4d9afb8bf55d614e962c5ed95

craffel:
Yes, I can.  What I'm trying to get you to do is rebase your fork so that it doesn't include this commit: https://github.com/rafaelvalle/pretty-midi/commit/fceaf52dbd265fe2cc194cb66d082766a833a42a and so that all commits come after https://github.com/craffel/pretty-midi/commit/30abab22f4ebcc136fa0fba6aa763b5bd1e07879 rather than some before and some after.  If you're having trouble doing this, the simplest way to do it would be to discard all of the commits you have here and create a new commit which applies the changes from all of these commits on top of HEAD of the main repository.

rafaelvalle:
Hey Collin. It is rebased now.

craffel:
Looks good, thanks.  I will squash with some minor edits and merge later today.

rafaelvalle:
Sounds good to me. I will wait before adding the second PR with get_histogram and get_transition_matrix

craffel:
I'm looking over this and trying it out again, and I think a better way to handle not including pitch bends in piano roll and chroma would just be for the user to manually remove (potentially temporarily) the pitch bends before calling `get_piano_roll` or `get_chroma`, i.e. you can achieve the same behavior with one or two lines in user code.  So, I think I'm not sure it warrants its own keyword argument, and I'm wary of including it because the piano roll/chromagram will be arguably incorrect when pitch bends aren't included.   Since it's disjoint from this PR, I am going to remove it before merging, we can talk about whether we want to include it in another PR later.

craffel:
Merged in a53537b5825efd34b8210cb958c919efc166421f.

PR #42, opened 2015-08-12
Conversion methods, layout and write methods

rafaelvalle:
Creating methods for conversion
Updating pretty_midi to properly write midi files with key and time signatures
Fixing layout


craffel:
Merged.

PR #43, opened 2015-08-12
PC Histogram and PC Transition Matrix

rafaelvalle:
Adding routines to compute pitch class histogram and pitch class transition matrix.


craffel:
I don't think your rebases are working. Note how there are 40 commits listed here, when really there have only been two or so commits since your last change.  What does your local log look like?  It shouldn't include any of the commits which are different from the main fork except for two two that you've made since my last merge.  This was true in the last PR too but I just cherry-picked.  I can look at this with you in person too.

rafaelvalle:
Let's look at in in person as I might be doing the rebase wrong!

PR #44, opened 2015-08-12
adding get_pc_histogram and pc_transition_matrix

rafaelvalle:
Adding methods for calculating pitch class histogram and transition matrix.

### DIFF
--- a/pretty_midi/instrument.py
+++ b/pretty_midi/instrument.py
@@ -15,11 +15,6 @@ from .utilities import pitch_bend_to_semitones, note_number_to_hz
 
 DEFAULT_SF2 = 'TimGM6mb.sf2'
 
-def set_trace():
-    from IPython.core.debugger import Pdb
-    import sys
-    Pdb(color_scheme='Linux').set_trace(sys._getframe().f_back)
-
 
 class Instrument(object):
     """Object to hold event information for a single instrument
@@ -250,16 +245,14 @@ class Instrument(object):
 
         return histogram
 
-    def get_pitch_class_transition_matrix(self, use_duration=False,
-                                          normalize=False):
-        """Computes the pitch class transition matrix of the current instrument,
-        optionally weighted by their durations.
+    def get_pitch_class_transition_matrix(self, normalize=False):
+        """Computes the pitch class transition matrix of the current instrument
+
+        Transitions are added whenever the end of a note is within 50ms from the
+        start of any other note.
 
         Parameters
         ----------
-        use_duration : bool
-            Increase frequency by transition duration (current and
-            next note)
         normalize : bool
             Normalize transition matrix such that matrix sum equals is 1.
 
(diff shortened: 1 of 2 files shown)


craffel:
Thanks, these look good.  Some comments - 

1. I think we should put them in `Instrument`, and have `PrettyMIDI` call its instruments, like the way `get_piano_roll` works.  `PrettyMIDI`'s method would just call the corresponding methods with `normalize=False` for each of its instruments, sum them up, and then optionally normalize.  I mainly say this because there are applications where it would be useful to get the histogram/transition matrix for individual instruments.

1. Shouldn't they exclude drum instruments?  `Instrument.get_piano_roll` does this by just returning zeros when it is a drum instrument.

1. Wouldn't it be useful to be able to weight by velocity, too?

1. I'm suspicious `np.histogram` is faster, e.g. for an instrument method:
```
np.histogram([n.pitch % 12 for n in self.notes],
             bins=np.arange(12),
             weights=[n.end - n.start for n in self.notes])
```

rafaelvalle:
1. That's a good idea. I will update it.
2. Exactly, will update as well!
3. Having the option would be useful. I'm not sure, however, how import velocity is and what it's weight should be.
4. Yes, np.histogram is twice as fast! Will update it.

craffel:
Great.

> Having the option would be useful. I'm not sure, however, how import velocity is and what it's weight should be.

I think the only sane way to do it would be to weight according to the raw velocity.  Something like

```Python
weights = np.ones(len(self.notes))
if use_duration:
    weights *= [note.end  - note.start for note in self.notes]
if use_velocity:
    weights *= [note.veocity for note in self.notes]
```

craffel:
A few suggested changes -

1. Instead of initializing "empty" results to dtype int, don't specify a dtype (which will result in float).  This makes it consistent with the "non-empty" dtype.

1. Have the same argument defaults in the `Instrument` and `PrettyMIDI` versions of these functions.  I think they should all be False in all cases for all parameters, just for simplicity.

1. Handle chords/non-sorted notes when constructing the pitch-class transition matrix.  I'll let you figure this one out :)

pretty_midi/instrument.py
```diff
@@ -15,6 +15,11 @@
 
 DEFAULT_SF2 = 'TimGM6mb.sf2'
 
+def set_trace():
```
craffel: Get this outta here!

pretty_midi/instrument.py
```diff
@@ -205,6 +210,163 @@ def get_end_time(self):
...
+        normalize : bool
+            Normalize transition matrix such that matrix sum equals is 1.
+
+        Returns
+        -------
+        pitch_class_transition_matrix : np.ndarray, shape=(12,12)
+            Pitch class transition matrix
+        """
+
+        # track is drum or just one note, return all zeros transition matrix
```
craffel: *instrument is drum.  Instrument != track

pretty_midi/instrument.py
```diff
@@ -205,6 +210,163 @@ def get_end_time(self):
...
+        if self.is_drum or len(self.notes) == 1:
```
craffel: len(self.notes) can be 0 too, so it should be `len(self.notes) <= 1`.

pretty_midi/instrument.py
```diff
@@ -205,6 +210,163 @@ def get_end_time(self):
...
+            return np.zeros((12, 12))
+
+        # container for holding pitch class transitions
+        pitch_class_transition_matrix = np.zeros((12, 12))
+
+        # self.notes is not sorted, sort them
+        sorted_notes = sorted(self.notes, key=lambda x: x.start)
+
+        # use 0.06s as the minimun time threshold for separate notes
+        time_thresh = 0.06
```
craffel: I used 50 ms (period of 20 Hz) for the separation between onsets which denotes "rhythmic information".  Why did you choose 60 ms?

pretty_midi/instrument.py
```diff
@@ -205,6 +210,163 @@ def get_end_time(self):
...
+
+        histogram, _ = np.histogram([n.pitch % 12 for n in self.notes],
+                                    bins=np.arange(13),
+                                    weights=weights,
+                                    density=normalize)
+
+        return histogram
+
+    def get_pitch_class_transition_matrix(self, use_duration=False,
+                                          normalize=False):
```
craffel: It makes sense (to me) to be able to weight by velocity here, too.

pretty_midi/instrument.py
```diff
@@ -205,6 +210,163 @@ def get_end_time(self):
...
+            weights *= [note.velocity for note in self.notes]
+
+        histogram, _ = np.histogram([n.pitch % 12 for n in self.notes],
+                                    bins=np.arange(13),
+                                    weights=weights,
+                                    density=normalize)
+
+        return histogram
+
+    def get_pitch_class_transition_matrix(self, use_duration=False,
```
craffel: It doesn't look like you're using `use_duration` anywhere.

craffel:
I made a few minor comments.  I'm not totally clear on the rules you are using to say that one note has transitioned into another note, so we should talk Monday.  I think that we can construct it as a maximum (weighted) bipartite matching, which could potentially be more efficient and non-greedy.

pretty_midi/instrument.py
rafaelvalle: ~ 16 notes per sec. 20 notes per second is fine as well.

rafaelvalle:
I'm playing at Poisson Rouge on Monday and will work from home to save
time! Let's talk on Tuesday!

pretty_midi/instrument.py
rafaelvalle: Forgot to remove as it from the method declaration.

pretty_midi/instrument.py
rafaelvalle: Fixed!

pretty_midi/instrument.py
rafaelvalle: Exactly! Fixed!

pretty_midi/instrument.py
rafaelvalle: !

pretty_midi/instrument.py
```diff
@@ -205,6 +205,93 @@ def get_end_time(self):
...
+        # compute distance matrix for all end and start time pairs
+        dist_mat = np.subtract.outer(data_mat[:, 1], data_mat[:, 0])
+
+        # compute boolean matrix of indices within threshold
+        trans_idx= abs(dist_mat) < time_thresh
+
+        # update transition matrix
+        for i in xrange(len(data_mat)):
+            pitch_class_transition_matrix[
+                data_mat[i, 2], data_mat[trans_idx[i], 2].astype(int)] += 1
```
craffel: I like the idea here, but I don't think this will do what you want it to do; specifically, if you end up constructing multiple index tuples which are the same in one of these loop iterations, only 1 will get added to it.  For example,
```Python
>>> a = np.zeros((4, 4))
>>> a[[2, 2, 2, 2], [3, 3, 3, 3]] += 1
>>> print a

[[ 0.  0.  0.  0.]
 [ 0.  0.  0.  0.]
 [ 0.  0.  0.  1.]
 [ 0.  0.  0.  0.]]
```
I think what basically means that if note `i` transitions to the same pitch multiple times throughout the instrument, the pitch class transition matrix will only get one transition added.  You can confirm this by noting that `trans_idx.sum() != pitch_class_transition_matrix.sum()` (i.e. the number of notes whose start times are within the tolerance of another notes end time is not the same as the number of transitions recorded).

One way to do all this, which I think will be hard to beat in terms of efficiency, is with `np.histogram2d`:

```Python
trans_idx = np.argwhere(abs(dist_mat) < time_thresh)
pitch_class_transition_matrix = np.histogram2d(
    data_mat[trans_idx[:, 0], 2],
    data_mat[trans_idx[:, 1], 2], bins=12)[0]
```

Semantically, though, I like this the most:

```Python
# instrument is drum or less than one note, return all zeros
if self.is_drum or len(self.notes) <= 1:
    return np.zeros((12, 12))

# use 20hz(0.05s) as the minimun time threshold for separate notes
time_thresh = 0.05

# Retrieve note starts, ends, and transitions from self.notes
starts, ends, transitions = np.array(
    [[x.start, x.end, x.pitch % 12] for x in self.notes]).T

# Find indices of notes pairs of notes where the end time of
# one note is within time_thresh of the start time of the other
sources, targets = np.argwhere(abs(dist_mat) < time_thresh).T

return np.histogram2d(
    transitions[sources], transitions[targets], bins=12)[0]
```

rafaelvalle: Hard to beat this elegant and efficient version! Will copy paste !

craffel: > Hard to beat this elegant and efficient version! Will copy paste !
Note, however, that bins has to be np.arange(13).

Yes, very true!  Also, the `dist_mat` line got lost somewhere.  Should now be 
```Python
# compute distance matrix for all end and start time pairs
dist_mat = np.subtract.outer(starts, ends)
```

craffel: > "the output of argwhere is not suitable for indexing arrays". I'll substitute it for np.where

I hadn't seen this before, is that an error you're getting?  I believe `np.where` is the same as `np.nonzero` when you don't give it a condition argument... either one would actually be better than `np.argwhere` though, I think, because you can avoid the transpose!

rafaelvalle: http://docs.scipy.org/doc/numpy-dev/reference/generated/numpy.argwhere.html

craffel: Ah, they're just referring to the fact that you shouldn't use it directly to index an array (which we aren't doing).  In this use-case, of course it's suitable, because it works!

craffel:
Cool, looks good to me overall. One last question: Currently, the pitch class transition matrix is structured such that `pitch_class_transition_matrix[i, j]` is the number or proportion of transitions from pitch class `j` to pitch class `i`.  E.g.
```Python
>>> i = pretty_midi.Instrument(0, 0)
>>> i.notes.append(pretty_midi.Note(100, 0, 1., 3.))
>>> i.notes.append(pretty_midi.Note(100, 5, 3.01, 4.))
>>> i.get_pitch_class_transition_matrix(True)[5, 0]
1.0
```
Is this the standard way of doing it?  It would be a little more intuitive to me if it was transposed, so that `pitch_class_transition_matrix[i, j]` is the number or proportion of transitions from pitch class `i` to pitch class `j`.  This could be achieved by swapping the order of the first two arguments in `np.histogram2d` or by transposing the matrix.

craffel:
Looks good, thanks.  Merged in 211383ccd12e343e21b385b0821cde80d23aa7cb.  I fixed a few PEP-8s and simplified the methods in `PrettyMIDI` a bit, see https://github.com/craffel/pretty-midi/blob/master/pretty_midi/pretty_midi.py#L596

PR #45, opened 2015-08-27
adding routine for estimating beat times using metadata

rafaelvalle:
Using Time Signatures and Tempi to estimate beat locations.
For testing, I've created some .midi examples that address individual cases. Interestingly, floating point precision has been challenging to handle, as epsilon difference between beat locations must be treated as the same beat. Would be great to talk about this in person!

### DIFF
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -413,7 +413,7 @@ class PrettyMIDI(object):
         """
 
         # threshold in seconds for merging beat locations
-        epsilon = 0.05
+        epsilon = 0.01
 
         # get qpms (quarter per minute) tempi and their change time
         tempo_change_times, qpms = self.get_tempo_changes()
@@ -462,6 +462,7 @@ class PrettyMIDI(object):
                     cur_ts = self.time_signature_changes[ts_idx-1]
                     nxt_ts = self.time_signature_changes[ts_idx]
                 else:
+                    nxt_ts = None
                     break
 
             # update bpm given current time signature and qpm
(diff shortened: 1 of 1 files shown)


rafaelvalle:
Considered refactoring the code to have only one use case instead of four and concluded that it would make the code too complex.

craffel:
Can you explain why?  To me, having four separate cases makes the code very complex.

rafaelvalle:
Actually, the refactoring was quite simple after I tried a different approach. 
just pushed it.

pretty_midi/pretty_midi.py
```diff
@@ -402,6 +403,161 @@ def estimate_tempo(self):
...
+        if len(self.time_signature_changes) > 1:
+            cur_ts = self.time_signature_changes[0]
+            nxt_ts = self.time_signature_changes[1]
+        else:
+            cur_ts = self.time_signature_changes[0]
+            nxt_ts = None
+
+        # go over tempo change times and time signatures assuming that the first
+        # tempo change time is the start time of the first beat
+        i = 1
```
craffel: For consistency/readability it would be cool if `i` was renamed `tempo_idx` or something.  I did this wrong in `get_beats` and considered changing it but haven't yet... now would not be a bad time!

pretty_midi/pretty_midi.py
```diff
@@ -402,6 +403,161 @@ def estimate_tempo(self):
...
+            # time signature happens before the next tempo change. note that
+            # time signatures always begin on the start of a beat
+            while nxt_ts is not None and nxt_ts.time < tempo_change_times[i]:
+                # update bpm given current time signature and qpm
+                bpm = qpm_to_bpm(qpms[i-1],
+                                 cur_ts.numerator,
+                                 cur_ts.denominator)
+
+                # interpolate from cur to nxt time signature change time
+                beat_times = np.hstack((beat_times, np.arange(beat_start,
```
craffel: Don't forget, `np.arange` is bad with non-integer step sizes!  Use `np.linspace` instead.  Also, much more of a minor point: `np.append` is a little more readable (to me) with 1d arrays.

pretty_midi/pretty_midi.py
```diff
@@ -402,6 +403,161 @@ def estimate_tempo(self):
         """
         return self.estimate_tempi()[0][0]
 
+    def get_beats_using_metadata(self):
```
craffel: As we discussed, let's replace `get_beats` with this instead of making a new function.

craffel:
Looking good.  I made a few minor comments.  Did you also verify that this does the right thing in songs where a tempo change occurs between beats?  And did you check that this and the current `get_beats` return the same thing when there is only one time signature, 4/4 at 0s?  I can send some example MIDIs to check for that latter point.

pretty_midi/pretty_midi.py
rafaelvalle: Whenever a tempo change falls within a beat, one has to first find where the start of that beat is such that it can be used as the end time of np.linspace and to find out the number of beats. 
np.arange provides this end time and the beats in between as well.
I'm post processing the beat array to remove wrong beat times that np.arange might produce.

0 2.5

craffel: What do you mean?  `np.linspace` and `np.arange` can be used equivalently, always, except that linspace is never wrong.  If you know the start, stop, and step size, as you do here:

```Python
In [1]: start = 0.2
In [2]: stop = 2.2
In [3]: step = .1
In [4]: np.arange(start, stop + step, step)
Out[4]:
array([ 0.2,  0.3,  0.4,  0.5,  0.6,  0.7,  0.8,  0.9,  1. ,  1.1,  1.2,
        1.3,  1.4,  1.5,  1.6,  1.7,  1.8,  1.9,  2. ,  2.1,  2.2]

In [5]: np.linspace(start, stop, (stop - start)/step + 1)
Out[5]:
array([ 0.2,  0.3,  0.4,  0.5,  0.6,  0.7,  0.8,  0.9,  1. ,  1.1,  1.2,
        1.3,  1.4,  1.5,  1.6,  1.7,  1.8,  1.9,  2. ,  2.1,  2.2])
```

I'd prefer this a lot more than to allowing `np.arange` to potentially make unpredictable mistakes and then trying to fix them later.  That allows for a lot of edge cases that we may or may not account for.

rafaelvalle:
```python
nxt_tempo_change = 2.2
start = 0.25
step = 0.25

np.arange(start, nxt_tempo_change, step)

stop = start
while stop < nxt_tempo_change:
  stop += step
stop -= step

np.linspace(start, stop, (stop-start)/ step + 1)
```

rafaelvalle:
I've compared results with get_beats with 4/4 time signatures at 0s with and without tempo changes within beats. With tempo changes, sometimes get_beats produces one extra beat(last) and the results are the same without beat changes.

pretty_midi/pretty_midi.py
craffel: Ah, I see, so you're using np.arange as "accumulate until you have equalled or exceeded this point".  What do you want it to do when `start + n*step` is epsilon-close to `nxt_tempo_change`?  I'd rather this be handled explicitly, because `np.arange` is a good way to not be explicit about it.  For example, to ensure that `nxt_tempo_change` is never included, you could do something like 
```Python
new_beats = [beat_start]
while (new_beats[-1] < nxt_tempo_change and
       not np.allclose(new_beats[-1], nxt_tempo_change)):
    new_beats.append(new_beats[-1] + 60./bpm)
new_beats = new_beats[-1]
beat_times = np.append(beat_times, new_beats)
```

craffel:
> I've compared results with get_beats with 4/4 time signatures at 0s with and without tempo changes within beats. With tempo changes, sometimes get_beats produces one extra beat(last) and the results are the same without beat changes.

Great.  I have a few other things I need to finish up on, and then I'll run it through some tests and merge.

rafaelvalle:
Great! I'll push the key estimation code next!

craffel:
Cool, code looks good to me.  I am just now trying this out (sorry again for the delay), and it looks like it can get messy for some real-world MIDI files.  For example, the Bohemian Rhapsody.mid file included in the source distribution ends up having over 7,000 beats.  About a minute from the end, a ton of beats get reported by the new `get_beats`, here's a plot of beat index vs. beat time:

beat_starts

I was thinking this might be because there are tons of time signature change events for some weird reason, but `len(time_signature_changes) = 12`.  I haven't figured out why it's happening.  Do you mind taking a look?

craffel:
Hey Rafael, sorry for the delay, I'm finally getting back to this after deadlines and moving.

I tried out the latest code on Bohemian Rhapsody again, and while there isn't the strange issue where there are tons of beats at the end, the beats towards the end are instead very slow and almost certainly in the wrong places.  I spent some time looking through your code to try to figure out why but couldn't, so instead I just added what I understand to be your changes (use time signatures to convert qpm to bpm and reset beat location when a time signature is encountered) to the current `get_beats` and it appears to work fine.  I created a separate PR for this #48.  Can you test it against the "unit test" MIDIs you created to test this implementation, or send me the MIDIs?

rafaelvalle:
Hey Colin! Congrats on finishing these tasks.
The midi files are in my computer, which someone at JFK mistakenly took thinking it was theirs.
The code looks beautiful and I try the MIDI files next week, when I'm supposed to have my computer back!

craffel:
Ok sick, thanks Rafael!

rafaelvalle:
I tested it with the midi files and it turns out that there's a small bug in https://github.com/craffel/pretty-midi/blob/beats_with_time_signature/pretty_midi/pretty_midi.py#L448
Whenever there's a change in time signature and the last beat is marginally different from the new time signature's time, a repeated value is added to the list of beats

We could replace that condition with the code below.
I checked the method with the 'unittest' midis and with the Bohemian Rhapsody and everything looks correct

```python
if (ts_idx < len(self.time_signature_changes) - 1 and
    next_beat > self.time_signature_changes[ts_idx + 1].time):
    # Update the time signature index
    ts_idx += 1
    bpm = qpm_to_bpm(
        tempi[tempo_idx],
        self.time_signature_changes[ts_idx].numerator,
        self.time_signature_changes[ts_idx].denominator)
    # if last beat time is close to time signature time, replace
    # last beat and update next_beat according to new time signature
    # otherwise update next_beat
    if np.isclose(self.time_signature_changes[ts_idx].time,
                    beats[-1]):
        beats[-1] = self.time_signature_changes[ts_idx].time
        next_beat += beats[-1] + 60.0/bpm
    else:
        # Set the next beat to the time signature change time
        next_beat = self.time_signature_changes[ts_idx].time
```

craffel:
Resolved in 2b0efcf4b3833c22ce24758ce5ad368ff6b2bab8

PR #48, opened 2015-09-30
Update get_beats to use time signatures

craffel:
This addresses the changes in #45, except it instead modifies the existing `get_beats` function.  Namely, bpm is computed from qpm using the most recently seen time signature, and when a time signature event is found, it is used as the next beat location.

I think an ideal test would be something like the following:

Time signature 4/4 and tempo 120 at 0s
Tempo change at .8s to 100 qpm
Time signature change at 1.7s to 6/8
Time signature change at 2.8s to 4/4

The beats should then be 
```
[0.,
 0. + 60/120 = .5,
 .5 + .6*(60/120) + .4*(60/100) = 1.04,
 1.04 + 60/100 = 1.64,
 1.7,
 1.7 + 60/66.66666... = 2.6,
 2.8,
 2.8 + 60/100 = 3.4, ...]
```

### DIFF
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -14,6 +14,7 @@ from .containers import KeySignature, TimeSignature
 from .containers import Note, PitchBend, ControlChange
 from .utilities import mode_accidentals_to_key_number
 from .utilities import key_number_to_mode_accidentals
+from .utilities import qpm_to_bpm
 
 # The largest we'd ever expect a tick to be
 MAX_TICK = 1e7
@@ -423,39 +424,72 @@ class PrettyMIDI(object):
         # Create beat list; first beat is at first onset
         beats = [start_time]
         # Index of the tempo we're using
-        n = 0
+        tempo_idx = 0
         # Move past all the tempo changes up to the supplied start time
-        while (n < tempo_change_times.shape[0] - 1 and
-                beats[-1] > tempo_change_times[n]):
-            n += 1
+        while (tempo_idx < tempo_change_times.shape[0] - 1 and
+                beats[-1] > tempo_change_times[tempo_idx]):
+            tempo_idx += 1
+        # Index of the time signature change we're using
+        ts_idx = 0
+        # Move past all time signature changes up to the supplied start time
+        while (ts_idx < len(self.time_signature_changes) - 1 and
+                beats[-1] > self.time_signature_changes[ts_idx]):
+            ts_idx += 1
         # Get track end time
         end_time = self.get_end_time()
         # Add beats in
         while beats[-1] < end_time:
             # Compute expected beat location, one period later
-            next_beat = beats[-1] + 60.0/tempi[n]
+            bpm = qpm_to_bpm(tempi[tempo_idx],
+                             self.time_signature_changes[ts_idx].numerator,
+                             self.time_signature_changes[ts_idx].denominator)
+            next_beat = beats[-1] + 60.0/bpm
+            # If the next beat location passes a time signature change boundary
+            if ts_idx < len(self.time_signature_changes) - 1:
+                # Time of the next time signature change
+                next_ts_time = self.time_signature_changes[ts_idx + 1].time
+                if (next_beat > next_ts_time or
+                        np.isclose(next_beat, next_ts_time)):
+                    # Set the next beat to the time signature change time
+                    next_beat = self.time_signature_changes[ts_idx + 1].time
+                    # Update the time signature index
+                    ts_idx += 1
+                    bpm = qpm_to_bpm(
+                        tempi[tempo_idx],
+                        self.time_signature_changes[ts_idx].numerator,
+                        self.time_signature_changes[ts_idx].denominator)
             # If the beat location passes a tempo change boundary...
-            if (n < tempo_change_times.shape[0] - 1 and
-                    next_beat > tempo_change_times[n + 1]):
+            if (tempo_idx < tempo_change_times.shape[0] - 1 and
+                    next_beat > tempo_change_times[tempo_idx + 1]):
                 # Start by setting the beat location to the current beat...
                 next_beat = beats[-1]
                 # with the entire beat remaining
                 beat_remaining = 1.0
                 # While a beat with the current tempo would pass a tempo
                 # change boundary...
-                while (n < tempo_change_times.shape[0] - 1 and
-                        next_beat + beat_remaining*60.0/tempi[n] >=
-                        tempo_change_times[n + 1]):
+                while (tempo_idx < tempo_change_times.shape[0] - 1 and
+                        next_beat + beat_remaining*60.0/bpm >=
+                        tempo_change_times[tempo_idx + 1]):
+                    # Compute bpm adjusted for time signature
+                    bpm = qpm_to_bpm(
+                        tempi[tempo_idx],
+                        self.time_signature_changes[ts_idx].numerator,
+                        self.time_signature_changes[ts_idx].denominator)
                     # Compute the amount the beat location overshoots
-                    overshot_ratio = (tempo_change_times[n + 1] -
-                                      next_beat)/(60.0/tempi[n])
+                    overshot_ratio = (tempo_change_times[tempo_idx + 1] -
+                                      next_beat)/(60.0/bpm)
                     # Add in the amount of the beat during this tempo
-                    next_beat += overshot_ratio*60.0/tempi[n]
+                    next_beat += overshot_ratio*60.0/bpm
                     # Less of the beat remains now
                     beat_remaining -= overshot_ratio
                     # Increment the tempo index
-                    n = n + 1
-                next_beat += beat_remaining*60./tempi[n]
+                    tempo_idx = tempo_idx + 1
+                # Compute bpm adjusted for time signature
+                bpm = qpm_to_bpm(
+                    tempi[tempo_idx],
+                    self.time_signature_changes[ts_idx].numerator,
+                    self.time_signature_changes[ts_idx].denominator)
+                next_beat += beat_remaining*60./bpm
             beats.append(next_beat)
         # The last beat will pass the end_time barrier, so don't include it
         beats = np.array(beats[:-1])


craffel:
Here's an example which covers the above test:

```Python
import midi

def time_diff_to_tick(time, resolution, tempo):
    return int(time/(60./float(tempo)/float(resolution)))

# Instantiate a MIDI Pattern (contains a list of tracks)
pattern = midi.Pattern()
# Instantiate a MIDI Track (contains a list of MIDI events)
track = midi.Track()
# Append the track to the pattern
pattern.append(track)

# Tempo 120 at 0s
tempo = midi.SetTempoEvent(tick=0)
tempo.set_bpm(120)
track.append(tempo)

# Time signature 4/4 at 0s
time = midi.TimeSignatureEvent(tick=0)
time.set_numerator(4)
time.set_denominator(4)
track.append(time)

# Tempo change at .8s to 100 qpm
tempo = midi.SetTempoEvent(
    tick=time_diff_to_tick(.8, pattern.resolution, 120))
tempo.set_bpm(100)
track.append(tempo)

# Time signature change at 1.7s to 6/8
time = midi.TimeSignatureEvent(
    tick=time_diff_to_tick(.9, pattern.resolution, 100))
time.set_numerator(6)
time.set_denominator(8)
track.append(time)

# Time signature change at 2.8s to 4/4
time = midi.TimeSignatureEvent(
    tick=time_diff_to_tick(1.1, pattern.resolution, 100))
time.set_numerator(4)
time.set_denominator(4)
track.append(time)

# Add a long note event to force pretty_midi to report lots of beats
on = midi.NoteOnEvent(tick=100, velocity=20, pitch=midi.G_3)
track.append(on)
off = midi.NoteOffEvent(tick=1000, pitch=midi.G_3)
track.append(off)

# Save the MIDI file
midi.write_midifile("example.mid", pattern)

pm = pretty_midi.PrettyMIDI('example.mid')
print pm.get_tempo_changes()
print pm.time_signature_changes
print pm.get_beats()
```
yields

```
(array([ 0. ,  0.8]), array([ 120.,  100.]))

[TimeSignature(numerator=4, denominator=4, time=0.0),
 TimeSignature(numerator=6, denominator=8, time=1.7),
 TimeSignature(numerator=4, denominator=4, time=2.79909090909)]

[ 0.          0.5         1.04        1.64        1.7         2.6
  2.79909091  3.39909091  3.99909091  4.59909091  5.19909091  5.79909091]
```

Close enough for jazz!

pretty_midi/pretty_midi.py
```diff
@@ -423,39 +424,69 @@ def get_beats(self, start_time=0.):
...
+            if (tempo_idx < tempo_change_times.shape[0] - 1 and
+                    next_beat > tempo_change_times[tempo_idx + 1]):
                 # Start by setting the beat location to the current beat...
                 next_beat = beats[-1]
                 # with the entire beat remaining
                 beat_remaining = 1.0
                 # While a beat with the current tempo would pass a tempo
                 # change boundary...
-                while (n < tempo_change_times.shape[0] - 1 and
-                        next_beat + beat_remaining*60.0/tempi[n] >=
```
rafaelvalle: I will look into this, as this tempi is in qpm not bpm.

craffel:
In https://github.com/craffel/pretty-midi/pull/45#issuecomment-148839792, @rafaelvalle wrote

> I tested it with the 'unit test' midi files and it turns out that there's a small bug in https://github.com/craffel/pretty-midi/blob/beats_with_time_signature/pretty_midi/pretty_midi.py#L448
Whenever the next_beat passes a new time signature and the last beat is marginally different from the new time signature's time, a repeated value is added to the list of beats

> We could replace that condition with the code below.
I checked the method with the 'unittest' midis and with the Bohemian Rhapsody and everything looks correct
> ```Python
if (ts_idx < len(self.time_signature_changes) - 1 and
    next_beat > self.time_signature_changes[ts_idx + 1].time):
    # Update the time signature index
    ts_idx += 1
    bpm = qpm_to_bpm(
        tempi[tempo_idx],
        self.time_signature_changes[ts_idx].numerator,
        self.time_signature_changes[ts_idx].denominator)
    # if last beat time is close to time signature time, replace
    # last beat and update next_beat according to new time signature
    # otherwise update next_beat
    if np.isclose(self.time_signature_changes[ts_idx].time,
                    beats[-1]):
        beats[-1] = self.time_signature_changes[ts_idx].time
        next_beat += beats[-1] + 60.0/bpm
    else:
        # Set the next beat to the time signature change time
        next_beat = self.time_signature_changes[ts_idx].time

Ah, good catch.  Modifying my test-case from above, I can get a "double" beat near 4 seconds by adding 
```Python
# Time signature change at 4.0s to 3/4
time = midi.TimeSignatureEvent(
    tick=time_diff_to_tick(1.2, pattern.resolution, 100) + 1)
time.set_numerator(3)
time.set_denominator(4)
track.append(time)
```
after the last time signature change.  If I don't add the `+ 1` to the tick of the time signature event, I don't get a double beat.  Then, `get_beats` gives 
```
[ 0.          0.5         1.04        1.64        1.7         2.6
  2.79909091  3.39909091  3.99909091  4.00181818  4.60181818  5.20181818
  5.80181818  6.40181818  7.00181818]
```
so there's a beat at 3.99909091 then at 4.00181818, which differ by 0.00272727.  It doesn't look like this is small enough for `np.isclose` to catch, so I think we need a more generous threshold.  One option would be to use the standard 50 milliseconds.  Another option would be to just let the user prune these beats after the fact.  What do you think?  Do you have another example which produces beats which are closer than that?

rafaelvalle:
In the example where you don't add the +1 there are two double beats [3.39909091,  3.99909091,  4.00181818]. 
How about using a fixed and relative threshold ? np.isclose have both absolute and relative distances.

The examples I tested have beats that are either equal or less than 1-e14 different.
I'll send you the midi files over e-mail.

craffel:
> In the example where you don't add the +1 there are two double beats [3.39909091, 3.99909091, 4.00181818]. 
How about using a fixed and relative threshold ? np.isclose have both absolute and relative distances.

Hm, I don't get double beats with
```Python
# Time signature change at 4.0s to 3/4
time = midi.TimeSignatureEvent(
    tick=time_diff_to_tick(1.2, pattern.resolution, 100))
time.set_numerator(3)
time.set_denominator(4)
track.append(time)
```

```[ 0.          0.5         1.04        1.64        1.7         2.6
  2.79909091  3.39909091  3.99909091  4.59909091  5.19909091  5.79909091
  6.39909091  6.99909091]
```

Maying we're using different versions of the code or our systems are different.  But yeah, if you can send me the test MIDIs, that would be cool.

rafaelvalle:
My bad, I read the numbers wrong 3.39909091 3.99909091

craffel:
> My bad, I read the numbers wrong 3.39909091 3.99909091

Ah, sure, but where did the `4.00181818` come from?

rafaelvalle:
Copying your message :
If I don't add the ```+ 1``` to the tick of the time signature event, I don't get a double beat. Then, ```get_beats``` gives
``` python
[ 0.          0.5         1.04        1.64        1.7         2.6
  2.79909091  3.39909091  3.99909091  4.00181818  4.60181818  5.20181818
  5.80181818  6.40181818  7.00181818]
```

rafaelvalle:
Also, the second comparison should be changed from > to >=
```
if (ts_idx < len(self.time_signature_changes) - 1 and
    next_beat > self.time_signature_changes[ts_idx + 1].time):
```

craffel:
> If I don't add the + 1 to the tick of the time signature event, I don't get a double beat. Then, get_beats gives

> ```[ 0.          0.5         1.04        1.64        1.7         2.6
  2.79909091  3.39909091  3.99909091  4.00181818  4.60181818  5.20181818
  5.80181818  6.40181818  7.00181818]```


Sorry, I'm confused - when you don't add +1, you don't get a double beat, but if that's what `get_beats` gives, there's a double beat there -- `3.99909091  4.00181818`?

rafaelvalle:
I'm just copying what you wrote.
https://github.com/craffel/pretty-midi/pull/48#issuecomment-149292644

craffel:
So, with the + 1 you get the same output that I got?  That's good.  I don't get a double beat without the + 1.

craffel:
Thanks for sending your unit test files; using a resolution of 1024 makes it easy to place a time signature change exactly on a beat which demonstrates this problem.  Here's a simple example:

```Python

resolution = 1024
# Instantiate a MIDI Pattern (contains a list of tracks)
pattern = midi.Pattern(resolution=resolution)
# Instantiate a MIDI Track (contains a list of MIDI events)
track = midi.Track()
# Append the track to the pattern
pattern.append(track)

# Tempo 120 at 0s
tempo = midi.SetTempoEvent(tick=0)
tempo.set_bpm(120)
track.append(tempo)

# Time signature 4/4 at 0s
time = midi.TimeSignatureEvent(tick=0)
time.set_numerator(4)
time.set_denominator(4)
track.append(time)

# Fake time signature change resolution*4
time = midi.TimeSignatureEvent(tick=resolution*4)
time.set_numerator(4)
time.set_denominator(4)
track.append(time)

# Add a long note event to force pretty_midi to report lots of beats
on = midi.NoteOnEvent(tick=100, velocity=20, pitch=midi.G_3)
track.append(on)
off = midi.NoteOffEvent(tick=1000, pitch=midi.G_3)
track.append(off)

# Save the MIDI file
midi.write_midifile("example.mid", pattern)

pm = pretty_midi.PrettyMIDI('example.mid')
print pm.get_beats()
```

yields `[ 0.   0.5  1.   1.5  2.   2.   2.5]`.  I think the `isclose` solution is good, and I will add a note about how `get_beats` can still produce beats which are pretty close together in some cases.  We can leave it to the user to filter those out if they desire.

rafaelvalle:
Cool! I'll issue a request for key extraction pretty soon!

craffel:
Ok, just updated it with `isclose`.  It matches your output on all of the unit tests.  Merging!

PR #65, opened 2016-05-05
Added stable sorting to pretty_midi file writer.

douglaseck:


douglaseck:
HI Colin. Trying to figure out Git. I don't know what you think about this change. I've got a C++ MIDI writer that I'm using for batch processing. I wanted it to create byte-identical midi files (at least for files in tests) to pretty_midi. But the sorting of elements with the same absolute time is not predictable. I made it predictable for note ons, pitch bends and control changes. This made it easy to verify that my code writes exactly the same bytes.

douglaseck:
Ok these are the four fixes I have checked in to my local version. I'd like to go ahead and have them fixed upstream and then import a more recent version.  Please let me know if these changes make sense.

craffel:
Thanks, this looks reasonable to me.  So what you are saying is that when sorting by tick and multiple events occur at the same tick, you want to be able to know exactly what order those events will appear? It seems like to do this completely, there also needs to be a hierarchy of event types (e.g. when comparing a control change to a note message, the control change should always go first, or something, etc.) and it also might be worth doing for all of the possible events (e.g. when comparing two key signature changes which occur at the same time, which is an edge case that might happen).  What do you think?

douglaseck:
Agreed! I just did what it took to get my test to pass :-)

craffel:
Any thoughts on what the sort order for the remaining cases should be?  Are you up for implementing that or do you want me to?

douglaseck:
I'll take a shot at it. It seems pretty straightforward.

douglaseck:
Please give this a look.

pretty_midi/pretty_midi.py
```diff
@@ -902,12 +902,57 @@ def write(self, filename):
...
+          secondary_sort = {
+              'Set Tempo': lambda(e): (1 * 256 * 256),
+              'Time Signature': lambda(e): (2 * 256 * 256),
+              'Key Signature': lambda(e): (3 * 256 * 256),
+              'Program Change': lambda(e): (4 * 256 * 256),
+              'Pitch Wheel': lambda(e): ((5 * 256 * 256) + e.pitch),
+              'Control Change': lambda(e): (
+                  (6 * 256 * 256) + (e.control * 256) + e.value),
+              'Note Off': lambda(e): (
+                  (7 * 256 * 256) + (e.pitch * 256),
```
craffel: Looks like there is a missing ) here.

craffel:
Great, thank you, this looks reasonable to me.  It looks like you have a syntax error though (just line-commented it).  I can fix it when I PEP-8ify your changes, or feel free to change it yourself.

douglaseck:
Please change. I copy-pasted from my version and missed the close.

craffel:
Merged in https://github.com/craffel/pretty-midi/compare/0ca39fd1b1558468a402f1cf8fa126ce4ed7e76a...2dd7e5306bf627e7acd9bd4115c80c2060f0267a .  Thanks again!

PR #67, opened 2016-06-01
Remove sorting code not defined in event_compare

adarob:
Colin, I think @douglaseck forgot this in his stable sorting change, which is causing problems in tensorflow/magenta.

### DIFF
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -374,7 +374,7 @@ class PrettyMIDI(object):
         tempi = np.zeros(len(self._tick_scales))
         for n, (tick, tick_scale) in enumerate(self._tick_scales):
             # Convert tick of this tempo change to time in seconds
-            tempo_change_times[n] = self.__tick_to_time[tick]
+            tempo_change_times[n] = self.tick_to_time(tick)
             # Convert tick scale to a tempo
             tempi[n] = 60.0/(tick_scale*self.resolution)
         return tempo_change_times, tempi


fredbertsch:
What's going on here?

craffel:
Thanks for pointing this out.  I think the issue is that this snippet you removed and the new stable sorting code are doing the opposite thing - the stable sort always puts a note on event after a note off event when they have the same tick, and this snippet always puts the note off after the note on.  The initial justification for putting note offs second is to avoid a stuck note (e.g. a note-on with no corresponding note-off).  I think the justification behind putting note-ons second is to prevent the artificial removal of a note.  Does that make sense/are you with me?

I think the actual correct thing to do would be to determine the ordering based on whether a note-on appears before or not.  For example, in this scenario:
```
O ------- O/F ------- F ...
```
where "O" is a note on and "F" is a note off and "O/F" is a note on and note off with the same tick (all notes having the same pitch of course), we should put the "F" first, because then we have two complete notes.  In this scenario:
```
O ----- F ------ O/F ------- O ------ F ---- ...
```
we should really just remove both the note on and the note off event with the same tick, because they correspond to a note which has zero duration.

Does this all make sense?  Also would be great to have @douglaseck's input.

douglaseck:
Yes. Your justification makes sense. It's a bit annoying to have to do a
stateful sort (tracking what note-ons are currently active).

Stuck notes are horrible.  I think the safest thing to do is to restore
Colin's sorting approach, but do it by deleting the snippet and changing
the sort function I wrote.

craffel:
In an offline conversation, @douglaseck pointed out that my reading of the old sorting code was wrong, and they indeed are actually both doing the same thing (putting note-offs first when note-on and note-off appear at the same time).  Sorry!  But, I think we all agree that stateful sorting to avoid stuck notes is smartest,  Right?

PR #68, opened 2016-07-04
remove parentheses around lambda function arguments

vug:
Replaced "lambda(e): ..." with "lambda e: ..." because it was generating "SyntaxError: invalid syntax" in Python 3.


craffel:
Thanks, merging.  FYI, `pretty_midi` is generally untested in Python 3 (though this will change soon/eventually).  If you find any other issues, feel free to report them.

PR #70, opened 2016-07-13
Improvement to adjust_times and get_downbeats

craffel:
This PR addresses the fact that `adjust_times` was originally doing the absolute most simple thing, i.e. interpolating notes and pitch/control change locations.  It was not adjusting any track-level event locations, nor was it adjusting the tempo map of the MIDI object (resulting in, for example, `get_beats()` not changing its output at all).  This PR addresses those issues:

- Added a `get_downbeats` function, which was something we needed anyways but is also necessary to determine where to place the first time signature change.
- Added adjustment of track-level events (key and time signature changes).  Key changes just need to have their time interpolated, but care must be taken to include the final key change which appears before the beginning of `original_times` at the start of the MIDI, after mapping.  The first time signature change needs to be placed at the first adjusted downbeat location so that `get_beats` and `get_downbeats` have the correct offset.
- Added code which adds tempo change events where the new time map dictates that the timing is changing.  This effectively makes it so that e.g. `get_beats` produces the correct output after calling `adjust_times`.  This turned out to be particularly hairy because in MIDI files tempo changes must occur on a fixed integer "tick" grid, whose spacing is determined by the tempo changes themselves, but I have come up with a solution which appears to do the right thing.
- Misc other minor improvements for clarity/better code.

Any comments/review by `adjust_times` users would be appreciated!  Testing has been done ad-hoc in a big IPython notebook, actual tests are forthcoming.

### DIFF
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -1010,6 +1010,81 @@ class PrettyMIDI(object):
                 self.time_signature_changes.insert(
                     0, TimeSignature(4, 4, adjusted_downbeats[0]))
 
+        # Finally, we will adjust and add tempo changes so that the
+        # tick-to-time mapping remains valid
+        # The first thing we need is to map original_times onto the existing
+        # quantized tick grid, because otherwise when we are re-creating tick
+        # scales below the rounding errors accumulate and result in a bad,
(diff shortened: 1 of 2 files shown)


craffel:
Tests added.  LGTM, merging!

PR #84, opened 2016-08-31
Index instrument note ons and offs by channel rather than by channel program

douglaseck:
I'm not 100% convinced this works. But it *seems* to work. Please give it a look.

### DIFF
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -242,7 +242,7 @@ class PrettyMIDI(object):
             # If we have already created an instrument for this program
             # number/track/channel, return it
             if (program, is_drum, channel, track) in instrument_map:
-                return instrument_map[(program, is_drum, channel, track)]
+              return instrument_map[(program, is_drum, channel, track)]
             # If there's a straggler instrument for this instrument and we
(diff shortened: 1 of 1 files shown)


douglaseck:
fixes #84 
Running the test code in the issue now yields:
[Instrument(program=0, is_drum=False, name=""), Instrument(program=0, is_drum=False, name="")]
[Note(start=0.022727, end=0.045455, pitch=43, velocity=20)]

I'm still not 100% sure this is a legit fix, but it makes sense to me.

craffel:
This seems like the right thing to me.  `current_instrument` is used to keep track of the most recent program change seen on each channel in a given track.  This is useful for determining which `Instrument` to associate with each completed note.  `last_note_on`, on the other hand, should be used for determining notes which appeared on a given track/channel/pitch prior to seeing a note-off.  So, we still need both, but it makes sense (after 0ca39fd1b1558468a402f1cf8fa126ce4ed7e76a) to use `event.channel` as the key rather than `current_instrument[event.channel]`.  I think this change was basically missing from 0ca39fd1b1558468a402f1cf8fa126ce4ed7e76a.

PR #87, opened 2016-09-01
Remove is_drum from _load_instruments

### LINKED ISSUE #86: `is_drum` is no longer necessary in `_load_instruments`
Since everything is being separated/keyed by channel, we no longer need to separately keep track of `is_drum` when indexing `note_on_index` and `__get_instrument`.

craffel:
Resolves #86.  Now that channel is used as a key for `note_on_index` and `__get_instrument`, we no longer need to separately key by `is_drum`, which is just defined as `channel == 9`.

@douglaseck can you check to see if this breaks any of your code?  There should be no change in behavior here.

### DIFF
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -233,7 +233,7 @@ class PrettyMIDI(object):
         # This dict will map track indices to any track names encountered
         track_name_map = collections.defaultdict(str)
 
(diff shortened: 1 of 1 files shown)


douglaseck:
This seems fine. Please go ahead and submit.

craffel:
> Please go ahead and submit.

Is this a genteel way of saying LGTM?  :)  Merging.

Issue #92, opened 2016-09-16
setup.py should install midi package from GitHub


craffel:
cc @cghawthorne

Some people have trouble with `pip install midi`, see #72 or https://github.com/vishnubob/python-midi/issues/66.   The `midi` package is also out-of-date (v0.2.3).  The `python-midi` package on PyPI is more up-to-date, but doesn't appear to be maintained by the owner of `midi`: https://github.com/vishnubob/python-midi/issues/72 And, despite being more up-to-date, as of writing it is over 2 years stale.  There are features/changes added since then which are needed in `pretty_midi`.  To deal with this, README suggests using the bleeding-edge from GitHub.  We should specify this in setup.py too.

craffel:
Resolved in #93.

craffel:
This is still causing problems.

craffel:
Problems are not resolvable without fixing `midi` on PyPI.

PR #93, opened 2016-09-16
Give GitHub link for midi package in setup.py

craffel:
Due to issues with installing midi from pip (see #92), this should explicitly make pip install the bleeding-edge.  @cghawthorne can you confirm this fixes the problem you were facing?


cghawthorne:
Works for me!

PR #95, opened 2016-09-26
Pull Python 3 compatible branch of python-midi instead

carlthome:
# Note: pull request not ready for merge.

# TODO
- [ ] Verify that python setup.py works with Python 2
- [ ] Verify that python setup.py works with Python 3

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,5 +27,5 @@ setup(
         'midi'
     ],
     dependency_links=[
-        'https://github.com/vishnubob/python-midi/tarball/master#egg=midi']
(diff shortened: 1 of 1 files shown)


carlthome:
Unfortunately python-midi's Python 3 branch is a bit behind master, so this would not be a sensible change... I'll try to get them to update master instead.

craffel:
I am in the process of changing the "backend" from `python-midi` to `mido`, because of PIP install issues and lack of maintaining on `python-midi`.  So, don't worry about this.

carlthome:
Nice! :smile:

PR #98, opened 2016-09-26
Minor Python 3 compatibility fixes

carlthome:
Minor compatibility issue with pm.write() fixed.

- In Python2, range returns a list.
- In Python3, range returns an iterator. 

Thus `channels` does not have a remove method in a Python 3 environment. This pull request fixes that by explicitly making `channels` into a list.


carlthome:
Unfortunately tests fail even on the master branch in both Python 2 and 3 (without any modifications on my part) so it's hard to tell if this minor change will cause any trouble (though I seriously doubt it).

carlthome:
Python 2.7 is the oldest version that has cmp_to_key. I think that should be okay, right?

craffel:
Hey, thanks for this.  There are a few other places where a comparison function is being used.  Do you want to put `cmp_to_key` on all of them in this PR?  

> Python 2.7 is the oldest version that has cmp_to_key. I think that should be okay, right?

Yes, requiring 2.7 is fine.

> EDIT: Perhaps adding Travis to this repo would be sweet? 

This will happen sometime.

craffel:
@carlthome now that `mido` is the backend, let's get this merged so that we can claim Python 3 compatibility!  Looks like you need to rebase.

carlthome:
Great news, @craffel!

Rebased and merged with current master. 👍 

I'm unable to find any other places using `cmp` though (searching through all .py files didn't pop up anything, what am I missing?)

craffel:
Oops, sorry, I misunderstood - thought the issue was whenever `key=` was used, not whenever `cmp=` was used.  You found the only usage of `cmp=`.  Merging!

Issue #34, opened 2015-02-18
MIDI files with corrupt pitch values are not handled correctly


craffel:
Sometimes, `midi` reads in a MIDI file with has `NoteOnEvents` with `data[0] > 127`:

```Python
In [1]: import midi
In [2]: m = midi.read_midifile("data/clean_midi/mid/Celine Dion/That's The Way It Is.mid")
In [3]: for t in m:
   ...:     for e in t:
   ...:         if type(e) == midi.NoteOnEvent:
   ...:            if e.data[0] > 127:
   ...:                print e
   ...:
midi.NoteOnEvent(tick=4656, channel=7, data=[253, 75])
midi.NoteOnEvent(tick=0, channel=7, data=[254, 75])
midi.NoteOnEvent(tick=48, channel=7, data=[253, 0])
midi.NoteOnEvent(tick=0, channel=7, data=[254, 0])
```

This is because `midi` just loads in data via `ord` https://github.com/vishnubob/python-midi/blob/master/src/fileio.py#L94, so as long as the argument is a 8-bit char it will happily set a data value to a number more than 127.  So, this likely happens for other events too, not just `NoteOnEvents` `data[0]`.  This can create issues later on.  Either `midi` should raise an exception, we should raise an exception, or we should issue a warning and ignore those events with invalid data values.

craffel:
This will be resolved by #99, because mido verifies that data bytes are less than 127: https://github.com/olemb/mido/blob/master/mido/midifiles.py#L282

craffel:
Resolved in 8fbbdec5904e0766ed14433e39379704fa87454e.

PR #101, opened 2016-10-03
Change backend parser to mido

### LINKED ISSUE #99: Change parser to mido
cc @douglaseck, @cghawthorne

The current backend, [`python-midi`](https://github.com/vishnubob/python-midi), does not appear very actively maintained.  The [last release was over 3 years ago](https://github.com/vishnubob/python-midi/releases), despite [63 commits](https://github.com/vishnubob/python-midi/compare/c66e93b529d4c65178c40a72096d26bd903743ac...HEAD) since then. The [current entry in PyPI](https://pypi.python.org/pypi/midi) is not installable (nothing has been uploaded).  There is [another entry in PyPI](https://pypi.python.org/pypi/python-midi), but it states that it is v0.2.4, which does not appear to be an official release, and either way is almost 2 years out of date.  The `python-midi` PyPI package also is not maintained by the actual package owner.   `pretty_midi` depends on commits to `python-midi` which have happened since v0.2.3, so `pretty_midi` is effectively not PIP-installable (not to mention the questionable nature of depending on a PyPI package which is not maintained by the package owner).  [There doesn't seem to be any activity](https://github.com/vishnubob/python-midi/issues/72#issuecomment-247674339) to solve these issues.

Alternatively, [`mido`](https://github.com/olemb/mido) also provides a pure-Python MIDI parser.  It was developed after `python-midi` was created, so using `mido` from the getgo was not an option.  It is much more actively maintained, with frequent releases, quick merging of PRs, etc.  Assuming there are no package-specific issues, we should switch to using `mido`.

craffel:
Resolves #99.  This commit changes the backend from the `python-midi` package to `mido`.  This change was tested by reading in 1,000 MIDI files with `pretty_midi` with both backends, writing them out, and then comparing md5 checksums of the written files.  This tests both parsing and writing, which are the two things the backend is used for.  The known differences are

- https://github.com/olemb/mido/issues/62; the two backends do different things on a tiny fraction of arguably invalid MIDI files.  This is arguably out of the scope of `pretty_midi`.
- #34 is implicitly resolved by this PR because mido appropriately raises when there are invalid data "bytes".
- `mido` rightfully rounds microseconds per quarter note in its BPM calculations, `midi` doesn't.  To maintain backwards compatibility, I am not using `mido`'s functionality for this, but I will soon.  This will break backwards compatibility (and change timing on the sub-microsecond scale).

@douglaseck, @cghawthorne can you run this code through your test suite to see if anything breaks before I merge and make a new release?

### DIFF
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -3,7 +3,7 @@ format
 
 """
 
-import midi
+import mido
 import numpy as np
 import math
 import warnings
@@ -11,11 +11,9 @@ import collections
 import copy
 
 from .instrument import Instrument
-from .containers import KeySignature, TimeSignature, Lyric
-from .containers import Note, PitchBend, ControlChange
-from .utilities import mode_accidentals_to_key_number
-from .utilities import key_number_to_mode_accidentals
-from .utilities import qpm_to_bpm
+from .containers import (KeySignature, TimeSignature, Lyric, Note,
+                         PitchBend, ControlChange)
+from .utilities import (key_name_to_key_number, qpm_to_bpm)
 
 # The largest we'd ever expect a tick to be
 MAX_TICK = 1e7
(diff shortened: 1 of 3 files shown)


pretty_midi/pretty_midi.py
```diff
@@ -49,19 +47,24 @@ def __init__(self, midi_file=None, resolution=220, initial_tempo=120.):
         """
         if midi_file is not None:
             # Load in the MIDI data using the midi module
-            midi_data = midi.read_midifile(midi_file)
+            midi_data = mido.MidiFile(midi_file)
```
cghawthorne: In order to continue support loading midi files from file objects, this should be something like:

```
            if type(midi_file) in (str, unicode):
              midi_data = mido.MidiFile(filename=midi_file)
            else:
              midi_data = mido.MidiFile(file=midi_file)
```

Which is similar to what the previous midi package did:
https://github.com/vishnubob/python-midi/blob/master/src/fileio.py#L158

cghawthorne: Thanks, verified that all our tests pass now!

PR #129, opened 2017-05-05
piano roll to pretty_midi object util

jsleep:
Created a basic utility function for myself and saw that it might be useful in the examples directory here. It doesn't do much but could be adapted for anyone's use.


craffel:
Thank you for this!  Before I do a detailed review, can you change it so that rather than being the utility function plus some globally-run test code, it's an actual runnable Python script? See https://github.com/craffel/pretty-midi/blob/master/examples/align_midi.py#L100 for an example.  In particular, the important part is to put anything not in the `piano_roll_to_pretty_midi` function inside a `if __name__ == '__main__':` block, so that the file can be imported without any side-effect code being run.  I think the best usage example would be one which took in an audio file, computed a CQT, did some thresholding, and converted that "piano roll" to a `pretty_midi.PrettyMIDI` object (i.e. a very crude transcription).  That's a pretty involved use-case, so maybe something simpler would be loading in a MIDI file, calling `get_piano_roll`, and then converting that piano roll back to a new `pretty_midi.PrettyMIDI` object.  Alternatively, if you want to just leave it as your simple test for now, that is OK too.

Separately, if you don't mind fleshing out the `piano_roll_to_pretty_midi` docstring (making it fully descriptive, in numpydoc style), that would be great.

After that, I can do a thorough code review, thanks!

jsleep:
@craffel just pushed some changes that do the simpler use case you requested - loading a MIDI file, converting it to piano roll, and then converting it back to a pretty midi object and writing that out to a new MIDI file. Doing the CQT case sounds interesting, I'll try and look into that as well.

craffel:
@carlthome do you mind looking this over too?

craffel:
Looking good - I will wait a day or two to see if @carlthome has any comments.

carlthome:
Looking now.

examples/reverse_pianoroll.py
```diff
@@ -0,0 +1,93 @@
...
+"""
+Utility function for converting a piano roll (integer matrix of shape
+(n_notes, n_frames)) to a pretty_midi.PrettyMIDI object. Note that this is a
+lossy process because certain information in a MIDI file (such as different
+instruments, pitch bends, overlapping notes, control changes, etc) cannot
+be stored in the piano roll matrix. To demonstrate the lossiness, this script
+includes a demonstration of parsing a MIDI file, constructing a piano roll
+matrix, and then creating a new MIDI from the piano roll.
+
+Most of this code is from @carlthome. @jsleep Adapted that code from mido
```
carlthome: Attribution not needed. Just adds documentation bloat IMHO.

examples/reverse_pianoroll.py
```diff
@@ -0,0 +1,93 @@
...
+        A pretty_midi.PrettyMIDI class instance describing
+        the piano roll.
+
+    '''
+    notes, frames = piano_roll.shape
+    pm = pretty_midi.PrettyMIDI()
+    instrument = pretty_midi.Instrument(program=program)
+
+    # pad 1 column of zeros so we can acknowledge inital and ending events
+    piano_roll = np.pad(piano_roll, [(0, 0), (1, 1)], 'constant')
```
carlthome: Nice. 👍

examples/reverse_pianoroll.py
```diff
@@ -0,0 +1,93 @@
...
+                    velocity=prev_velocities[note],
+                    pitch=note,
+                    start=note_on_time[note],
+                    end=time)
+            instrument.notes.append(pm_note)
+            prev_velocities[note] = 0
+    pm.instruments.append(instrument)
+    return pm
+
+if __name__ == '__main__':
```
carlthome: This looks like a unit test. I think the CLI with argparse should be removed, and that the translation from a example MIDI file to/from a piano roll should be refactored into a proper unit test run by the CI, with some assertion of a small enough reconstruction error (bends gone, velocity kept, some time shifting of events allowed, etc.)

craffel: This matches the structure of the other examples.  No need for a unit test on an example script.  The original script had a simple test in it (which I asked to be removed), so I am confident enough that it works.

carlthome: Confident, even further down the line after future modifications by other developers? Maybe run the scripts on example.mid in .travis.yml at least?

carlthome: Alright. Anyway, adding the examples to Travis should be a separate pull request.

E.g. something like:
`- python examples/reverse_pianoroll.py example.mid /tmp/example.mid` after [this](https://github.com/craffel/pretty-midi/blob/db1c3c3da98b4fa9bc8044d1e4e7566811cda1a3/.travis.yml#L20
).

I get how one would like to avoid running computationally expensive stuff in this manner though, but for these particular scripts it should be fine IMHO. :pancakes:

jsleep:
Fixed @carlthome 's requests other than the unit test. @craffel, I also have some code in the works for converting from audio to cqt to pretty_midi to midi, should I put that in a separate pull request?

carlthome:
@jsleep, I'd be very curious to see how you convert a CQT spectrogram to MIDI. I've been working full-time on that very problem for over a year now and regardless of using CQT, HPS, logmel or pure STFT, I can conclude it's a non-trivial task. OT of course, so feel free to email me.

craffel:
> I also have some code in the works for converting from audio to cqt to pretty_midi to midi, should I put that in a separate pull request?

How simple is it (maybe post a gist)?  Something really rudimentary (like, <4 lines of librosa code) which worked not-completely-horribly would actually be fun to include in this example, instead of my contrived midi-to-piano-roll-to-midi suggestion.

jsleep:
It's pretty sloppy code right now in a IPython notebook - but I'll create a gist right now just to show it.

jsleep:
Here's the gist: https://gist.github.com/jsleep/5d0c438124cfb59a4b0687d984e31606
I would put the cqt to piano roll in its own function. It creates a rough transcription - but seems like it may be an octave off for the monophonic track I tested it on.

carlthome:
I don't think thresholding a CQT spectrogram and interpreting values over the threshold as pitch activations is useful. It would almost never work, and in the case of monophonic signals you'd get much further in most cases by simply using YIN. :cry: 

[Here's](https://arxiv.org/abs/1612.05153) a good read if you're interested in advancing the state-of-the-art! :smiley: 

And [here's a notebook](https://gist.github.com/carlthome/1e7244e31bd628a0dba233b6dceebaef) with a time-frequency approach similar to yours but that could work with [missing fundamentals](https://en.wikipedia.org/wiki/Missing_fundamental) (but never with vocals or clarinets :rofl: ).

craffel:
@jsleep Cool, looks reasonable to me.  It could definitely be made more terse though - e.g. replace lines 16-28 with
```Python
piano_roll = np.abs(librosa.cqt(
    y, sr=sr, fmin=librosa.midi_to_hz(min_midi), n_bins=max_midi-min_midi))
piano_roll = np.digitize(piano_roll, np.linspace(piano_roll.min(), piano_roll.max(), 127))
piano_roll[piano_roll < threshold] = 0
```
Do you want to replace the existing midi-piano roll-midi with this, and maybe I can play around with it and see if I have any simple improvements?

@carlthome The point here is not to get something even vaguely close to state of the art, but simply to show a usage example of `pretty_midi`.  I think thresholding a CQT and getting a really shitty transcription is a totally fun and interesting example which can easily be expanded with more sophisticated stuff for motivated people.  Implementing something which actually works reasonably well is completely out of the scope of the example.

jsleep:
@carlthome Thanks for the resources - I am looking into doing AMT with deep learning and haven't seen that paper yet.

@craffel Sure I'll replace it and clean it up a bit.

jsleep:
@craffel could you review my last push? I just replaced the file with a new one converted audio to cqt to piano roll to MIDI.

examples/cqt2midi.py
```diff
@@ -0,0 +1,134 @@
...
+"""
+import pretty_midi
+import numpy as np
+import librosa
+import sys
+import argparse
+
+
+def cqt_to_piano_roll(cqt, min_midi, max_midi, threshold):
+    '''Convert a CQT transformation into a piano roll representation by
```
carlthome: "CQT transformation" is like saying "CD disc".

"CQT spectrogram" is better.

examples/cqt2midi.py
```diff
@@ -0,0 +1,134 @@
+from __future__ import division
+"""
+Utility function for converting an audio file
+to a pretty_midi.PrettyMIDI object. Note that this method is nowhere close
+to the state-of-the-art in automatic music transcription.
+This just serves as a fun example for rough
+transcription which can be exanded on for anyone motivated.
```
carlthome: "exanded"

examples/cqt2midi.py
```diff
@@ -0,0 +1,134 @@
...
+    # Set up command-line argument parsing
+    parser = argparse.ArgumentParser(
+        description='Transcribe Audio file to MIDI file',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+    parser.add_argument('input_audio', action='store',
+                        help='Path to the input Audio file')
+    parser.add_argument('output_midi', action='store',
+                        help='Path where the transcribed MIDI will be written')
+    parser.add_argument('--sr', default=22050, type=int, action='store',
```
craffel: I think it's OK to just force using librosa's default here for the sampling rate, because if it's changed from 22050 various parameters (hop size etc) of the CQT should also be changed.  Also simplifies things a little.

examples/cqt2midi.py
```diff
@@ -0,0 +1,134 @@
...
+    piano_roll = np.abs(cqt)
+    piano_roll = np.digitize(piano_roll, np.linspace(piano_roll.min(),
+                                                     piano_roll.max(), 127))
+    piano_roll[piano_roll < threshold] = 0
+    piano_roll = np.pad(piano_roll,
+                        [(128-max_midi, min_midi), (0, 0)], 'constant')
+    return piano_roll
+
+
+def piano_roll_to_pretty_midi(piano_roll, fs=100, program=0):
```
craffel: Nitpick, but can you switch the order of these two functions? It's `piano_roll_to_pretty_midi` that you are showcasing here, so makes sense to put it first.

examples/cqt2midi.py
```diff
@@ -0,0 +1,134 @@
+from __future__ import division
+"""
+Utility function for converting an audio file
+to a MIDI file. Note that this method is nowhere close
```
craffel: I think it is better to say "pretty_midi.PrettyMIDI object" here instead of "MIDI file" since that's really what you are demonstrating here; the CQT-to-MIDI is just a simple example of what you might do with this.  Know what I mean?  Same goes for the filename, in this case.

examples/cqt2midi.py
```diff
@@ -0,0 +1,134 @@
...
+    piano_roll : np.ndarray, shape=(128,frames), dtype=int
+        Piano roll representation on audio.
+
+    '''
+    piano_roll = np.abs(cqt)
+    piano_roll = np.digitize(piano_roll, np.linspace(piano_roll.min(),
+                                                     piano_roll.max(), 127))
+    piano_roll[piano_roll < threshold] = 0
+    piano_roll = np.pad(piano_roll,
+                        [(128-max_midi, min_midi), (0, 0)], 'constant')
```
craffel: Can you run pylint before submitting?  Should be `128 - max_midi`, for example.

examples/cqt2midi.py
```diff
@@ -0,0 +1,134 @@
...
+        # use time + 1 because of padding above
+        velocity = piano_roll[note, time + 1]
+        time = time / fs
+        if velocity > 0:
+            if prev_velocities[note] == 0:
+                note_on_time[note] = time
+                prev_velocities[note] = velocity
+        else:
+            pm_note = pretty_midi.Note(
+                    velocity=prev_velocities[note],
```
craffel: Same comment about pylint; these lines are overindented.

jsleep:
@craffel I fixed most of the stuff you mentioned other than maybe some of the pylint. I did the indentation based on the errors that pep8 gave me - fixing the pylint errors now fails pep8 checks and thus the travis check. Is there a config file you have for pep8 or pylint that I can follow?

craffel:
Haha sorry that's frustrating... best choice is to just follow Travis:
https://travis-ci.org/craffel/pretty-midi/jobs/232606300
It uses vanilla pep8.  Sorry to distract you with my pylint errors.

jsleep:
@craffel completely forgot to ping you again but the small pep8 changes have been fixed, hopefully this is the one!

craffel:
Right on, thank you for all of this!  Really happy to finally have this example in the repo!

Issue #131, opened 2017-07-18
Reading duplicate note-off events results in zero-duration notes


bzamecnik:
When the input MIDI file contains note-off events for already inactive notes (with note-off events given before) pretty-midi still creates note events with zero duration. Writing such notes then results in spurious long-running notes that were not present originally.

Note that in this case there were two overlapping notes: note-on, note-on, note-off, note-off.

Example: [bach-js_kunst_der_fuge_01_(c)unknown[2].mid](https://filebin.ca/3ThfIaDkkjvD/bach-js_kunst_der_fuge_01_cunknown2.mid) (source: kunstderfuge.com)

![image](https://user-images.githubusercontent.com/446124/28276344-d80f4b06-6b40-11e7-9078-2541f710987a.png)

If we zoom in the the start of the track 1 (first voice). Original looks like:

![image](https://user-images.githubusercontent.com/446124/28301674-a643cc56-6bb2-11e7-87df-517e0f4f7858.png)

And written one:

![image](https://user-images.githubusercontent.com/446124/28301685-affcb9c4-6bb2-11e7-9230-af3497be71b1.png)

Mido `print_track()` - original:

```
<meta message midi_port port=1 time=0>
<meta message track_name name='Contrapunctus I' time=0>
<message control_change channel=0 control=121 value=0 time=0>
<message note_on channel=0 note=69 velocity=88 time=1536>
<message note_on channel=0 note=74 velocity=88 time=192>
<message note_on channel=0 note=69 velocity=0 time=0>
<message note_on channel=0 note=72 velocity=88 time=192>
<message note_on channel=0 note=74 velocity=0 time=0>
<message note_on channel=0 note=69 velocity=88 time=192>
<message note_on channel=0 note=72 velocity=0 time=0>
<message note_on channel=0 note=68 velocity=88 time=192>
<message note_on channel=0 note=69 velocity=0 time=0>
<message note_on channel=0 note=69 velocity=88 time=192>
<message note_on channel=0 note=68 velocity=0 time=0>
<message note_on channel=0 note=71 velocity=88 time=96>
<message note_on channel=0 note=69 velocity=0 time=0>
<message note_on channel=0 note=72 velocity=88 time=96> <-- note-on
<message note_on channel=0 note=71 velocity=0 time=0>
<message note_on channel=0 note=72 velocity=88 time=192> <-- another note-on!
<message note_on channel=0 note=72 velocity=0 time=0> <-- proper note-off
[...]

Issue #52, opened 2015-11-25
Writing and reading back in should be roughly lossless


craffel:
I.e. if you write a MIDI file out and read it back in, you should get at least the same collection of notes, time signature changes, tempo changes, pitch bends, key signatures, etc. (all of the data that `pretty_midi` stores)

douglaseck:
I was going to file a bug but I'll attach it to this bug. In TimeSignature, you don't keep around the notated 32nd-notes in a MIDI quarter-note (24 MIDI Clocks) so when you write a file, this information is lost. An incoming TimeSignature might be:  
midi.TimeSignatureEvent(tick=0, data=[6, 3, 36, 8]). 
If you load that into pretty_midi and write it back out you get:
midi.TimeSignatureEvent(tick=0, data=[6, 3, 0, 0]),

If you don't want to store info, you could at least set some reasonable default such as you do when you write the default 4/4 time signature:
   timing_track += [midi.TimeSignatureEvent(tick=0, data=[4, 2, 24, 8])]
(24 midi clocks per metronome click, 8 32nd notes in a midi qn)

That seems like a reasonable default. I personally don't care all that much. I just don't want these files to break some sequencer software because of 0 values in these fields.

douglaseck:
Also it's probably not a good idea to always add a default time signature.  At the very least only add a time signature if there's not already a time signature defined at time 0.

craffel:
> In TimeSignature, you don't keep around the notated 32nd-notes in a MIDI quarter-note (24 MIDI Clocks) so when you write a file, this information is lost.

Ah, interesting, I hadn't noticed that we were missing that information when writing out ([here](https://github.com/craffel/pretty-midi/blob/master/pretty_midi/pretty_midi.py#L919)).  Do you know if this is totally meta-information, or will they ever effect playback?

> Also it's probably not a good idea to always add a default time signature. At the very least only add a time signature if there's not already a time signature defined at time 0.

I think I read somewhere that it's bad practice in a MIDI file to not include any time signature changes, but it turns out about 10% of the MIDI files I've found "in the wild" have no time signature events.  So, not sure which the best choice is, but I agree that we shouldn't add a default time signature at time 0 if there's already one there.

douglaseck:
I doubt if it affects playback. It broke a unit test that verifies that a
PM to proto to PM loop creates the same prettymidi info. Seems worthwhile
doing the minimally invasive fix for now. Those kinds of tests are useful
[...]

PR #132, opened 2017-07-18
#131 fix swapped note-on note-off events

bzamecnik:
This fixes the issue #131 (and helps to solve #52). Basically in case there are adjacent notes in the MIDI but at the same tick the note-on precedes note-off it's meaning was to close the previous note and open new note. The library however closed all open notes, including this. It created an empty event and sometimes long running notes that were closed only after encountering a matching note-off.

The solution was to keep the note-on event if it's at the same tick and we already closed some events.

I created a minimal test covering both cases - correct and swapped order. The input MIDI is produces via `mido`. We test both reading and writing.

Test are green and on real data piano rolls it seems the bug is fixed.

### DIFF
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -287,7 +287,7 @@ class PrettyMIDI(object):
         for track_idx, track in enumerate(midi_data.tracks):
             # Keep track of last note on location:
             # key = (instrument, note),
-            # value = (note on time, velocity)
+            # value = (note-on tick, velocity)
             last_note_on = collections.defaultdict(list)
             # Keep track of which instrument is playing in each channel
             # initialize to program 0 for all channels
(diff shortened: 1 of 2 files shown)


bzamecnik:
Possibly it might be good to write to a temporary directory and also to test whether the note duration is not by chance zero.

tests/test_pretty_midi.py
```diff
@@ -272,3 +273,59 @@ def simple():
...
+        pm_song = pretty_midi.PrettyMIDI(midi_path)
+
+        def extract_notes(pm_track):
+            return np.array([(note.pitch, note.end-note.start) for note
+                             in pm_track.notes])
+
+        expected = np.array([[72, 0.05], [72, 0.05], [74, 0.05], [72, 0.05]])
+        assert np.allclose(expected, extract_notes(pm_song.instruments[0]))
+
+        midi_written_path = 'overlapping_notes_%s_track_written.mid' % kind
```
craffel: Can you write this to a NamedTemporaryFile instead?

pretty_midi/pretty_midi.py
```diff
@@ -306,20 +306,37 @@ def __get_instrument(program, channel, track, create_new):
                     # Store this as the last note-on location
                     note_on_index = (event.channel, event.note)
                     last_note_on[note_on_index].append((
-                        self.__tick_to_time[event.time],
-                        event.velocity))
+                        event.time, event.velocity))
```
craffel: Any specific reason you are changing from time to ticks here?  I don't think floating point issues would bite us here unless they are nondeterministic :)

tests/test_pretty_midi.py
bzamecnik: Yes. Fixed in 83e8ebd. We reuse the NamedTemporaryFile while is open using `file.seek(0)`, otherwise on `close()` it gets deleted.

pretty_midi/pretty_midi.py
bzamecnik: Coming from computer graphics, comparing floats by equality scares me. We always had to use epsilon, otherwise nightmares may happen. When we have the original integers at hand it seems to me that it's safer to use them for comparison.

Issue #135, opened 2017-08-19
piano_roll_integrated error in class Instrument get_piano_roll method if fs >= 10


sccds:
`python 2.7.13`, `pretty_midi 0.2.8`

When we use `get_chroma()` method in class `PrettyMIDI`, if we set parameter `fs` >= 10, the chromagram matrix we get sometimes contains `NaN` column. And with `numpy` warning:
```
/usr/local/lib/python2.7/site-packages/numpy/core/_methods.py:59: RuntimeWarning: Mean of empty slice.
  warnings.warn("Mean of empty slice.", RuntimeWarning)
/usr/local/lib/python2.7/site-packages/numpy/core/_methods.py:68: RuntimeWarning: invalid value encountered in true_divide
  ret, rcount, out=ret, casting='unsafe', subok=False)
```

Traceback this issue, the problem is under `get_piano_roll()` method in class `Instrument`.  Here I show the last several lines above method `return`:
```python
if times is None:
            return piano_roll
piano_roll_integrated = np.zeros((128, times.shape[0]))
# Convert to column indices
times = np.array(times*fs, dtype=np.int)
for n, (start, end) in enumerate(zip(times[:-1], times[1:])):
    # Each column is the mean of the columns in piano_roll
    piano_roll_integrated[:, n] = np.mean(piano_roll[:, start:end],axis=1)
return piano_roll_integrated
```
When we compute `piano_roll_integrated`, we use `np.mean` of `start:end`. But if `start` and `end` are in same value, the `length` of `start:end` will be `zero`. 

Now lets look at why it is possible `start` and `end` value is possible to be same with parameter `fs` = 10. Here I abstract a small part of `numpy` test, but for this case, it is exactly true values in my project.
```python
import numpy as np
times = np.arange(5., 85, 0.1)
print times
times2 = np.array(times * 10, dtype=np.int)
print times2
print np.size(times), np.size(times2)
```
We can get the `times2` result is `[ 50  51  51  52  53  54  55  56  57  58 ...`. There are two `51`s in index 1 and index 2. Which cause problems.

This is because of the precision of number 52.0 is `51.999999999999993`, when using `dtype=np.int`, it ignores decimal part and equal to `51`.

To fix this issue, we can simply change it like this:
```python
import numpy as np
times = np.arange(5., 85, 0.1)
print times
times2 = np.array(np.round(times * 10), dtype=np.int)
print (times * 10)[2]
print times2
print np.size(times), np.size(times2)
```
[...]

PR #136, opened 2017-08-27
fixed issue #135

sccds:
fixed issue #135.
in `get_piano_roll` function, `start` and `end` index can be same because of float precision problem.
use `np.round()` to fix issue


craffel:
Thanks!

Issue #130, opened 2017-05-11
get_piano_roll does not take the MIDI sustain CC into account


craffel:
MIDI CC 64 is the "sustain pedal" control channel.  If it's on, notes should be held until it's turned off.  get_piano_roll doesn't take this into account.

maezawa-akira:
Hi, I added a flag to explicitly look at CC64.  It also contains a threshold pedal on/off, to allow handing of both standard MIDI pedal specs (CC64>64 is pedal on) versus continuous pedals (CC64>0 is soft pedal on).

PR #142, opened 2018-04-10
Add support for sustain pedal (pianoroll and chromagram)

maezawa-akira:
I addressed issue #130 , and added a functionality to take into account sustain pedals, when extracting the piano-roll and the chromagram.

### DIFF
--- a/pretty_midi/instrument.py
+++ b/pretty_midi/instrument.py
@@ -74,8 +74,7 @@ class Instrument(object):
         return np.sort(onsets)
 
     def get_piano_roll(self, fs=100, times=None,
-                       use_pedal=False,
-                       pedal_threshold=64):
+                       pedal_threshold=None):
         """Compute a piano roll matrix of this instrument.
 
         Parameters
@@ -86,16 +85,11 @@ class Instrument(object):
         times : np.ndarray
             Times of the start of each column in the piano roll.
             Default ``None`` which is ``np.arange(0, get_end_time(), 1./fs)``.
-        use_pedal : Boolean
-            Control Change 64 (Sustain pedal) is reflected as elongation of
-            notes.
-            Default is False, which does nothing.  If True, then CC64 value
-            greater than pedal_threshold will be treated as pedal on and
-            the rest as pedal off.
-        pedal_threshold : Int
-            The threshold value for treating CC64 message as elongation of a
-            note.
-            Default is ``64``
+        pedal_threshold : int
+            Value of control change 64 (sustain pedal) message that is above
+            this value is reflected as pedal-on.  Pedals will be reflected as
+            elongation of notes in the piano roll.
+            Default is None, which ignores CC64 messages.
 
         Returns
         -------
@@ -126,7 +120,7 @@ class Instrument(object):
                        int(note.start*fs):int(note.end*fs)] += note.velocity
 
         # Process sustain pedals
-        if use_pedal:
+        if pedal_threshold is not None:
             CC_SUSTAIN_PEDAL = 64
             time_pedal_on = 0
             is_pedal_on = False
@@ -198,8 +192,7 @@ class Instrument(object):
                                                   axis=1)
         return piano_roll_integrated
 
-    def get_chroma(self, fs=100, times=None, use_pedal=False,
-                   pedal_threshold=64):
+    def get_chroma(self, fs=100, times=None, pedal_threshold=None):
         """Get a sequence of chroma vectors from this instrument.
 
         Parameters
@@ -210,16 +203,11 @@ class Instrument(object):
         times : np.ndarray
             Times of the start of each column in the piano roll.
             Default ``None`` which is ``np.arange(0, get_end_time(), 1./fs)``.
-        use_pedal : Boolean
-            Control Change 64 (Sustain pedal) is reflected as elongation of
-            notes.
-            Default is False, which does nothing.  If True, then CC64 value
-            greater than pedal_threshold will be treated as pedal on and
-            the rest as pedal off.
-        pedal_threshold : Int
-            The threshold value for treating CC64 message as elongation of a
-            note.
-            Default is ``64``
+        pedal_threshold : int
+            Value of control change 64 (sustain pedal) message that is above
+            this value is reflected as pedal-on.  Pedals will be reflected as
+            elongation of notes in the chromagram.
+            Default is None, which ignores CC64 messages.
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


pretty_midi/instrument.py
```diff
@@ -113,6 +125,25 @@ def get_piano_roll(self, fs=100, times=None):
...
+                       if _e.number == CC_SUSTAIN_PEDAL]:
+                time_now = int(cc.time*fs)
+                is_current_pedal_on = (cc.value > pedal_threshold)
+                if not is_pedal_on and is_current_pedal_on:
+                    time_pedal_on = time_now
+                    is_pedal_on = True
+                elif is_pedal_on and not is_current_pedal_on:
+                    subpr = piano_roll[:, time_pedal_on:time_now]
+                    pedaled = np.minimum(subpr.cumsum(1),
+                                         subpr.max(1)[:, np.newaxis])
```
craffel: I'm a little confused by this operation.  Isn't `subpr.cumsum(1)` always less than `subpr.max(1)` given that subpr is positive (and it better be)? If you want to dress all of this code up with comments it wouldn't hurt!

pretty_midi/instrument.py
```diff
@@ -84,6 +86,16 @@ def get_piano_roll(self, fs=100, times=None):
         times : np.ndarray
             Times of the start of each column in the piano roll.
             Default ``None`` which is ``np.arange(0, get_end_time(), 1./fs)``.
+        use_pedal : Boolean
+            Control Change 64 (Sustain pedal) is reflected as elongation of
+            notes.
+            Default is False, which does nothing.  If True, then CC64 value
+            greater than pedal_threshold will be treated as pedal on and
+            the rest as pedal off.
+        pedal_threshold : Int
```
craffel: Is there any standard for this?  Like what does it mean for a MIDI file to have a sustain pedal CC value of 1?

maezawa-akira: Not really.  It depends more on how a synthesizer is implemented.  For example, piano-specific synthesizers interpret CC value of 1 to be a very soft pedal (like a Yamaha Disklavier), while most software synthesizers treat them as a no-pedal.  
So it's really dependent on the context of the MIDI data and the use case of this library's user.

maezawa-akira:
I have added documentation for readability.  Also, while documenting the code, I found a much simpler method for taking the pedal, so I modified the code as well.

pretty_midi/instrument.py
craffel: Ok, interesting.  Along these lines, I think the most common user preference would be for sustain pedal to be taken into consideration (corresponding to `use_pedal=True` in your code).  For that reason I'd suggest we unify `use_pedal` and `pedal_threshold` into a single argument, `pedal_threshold`, which defaults to 64.  If `pedal_threshold is None`, then we don't use the sustain pedal. What do you think?

As an aside, as a nitpick, please list types in the docstring literally, e.g. bool and int instead of Boolean and Int.

maezawa-akira:
Hi, I modified the code so that it uses a single parameter pedal_threshold to determine if CC64 should be used ( 93b94c0 )

craffel:
Ok, final suggestion: Should it default to `64` instead of `None`?  I would think people would, by default, want sustain pedal to be taken into consideration, even though it breaks backwards compatibility (which we're allowed to do since we're still pre-v1!)

pretty_midi/instrument.py
```diff
@@ -113,6 +120,29 @@ def get_piano_roll(self, fs=100, times=None):
...
 
+        # Process sustain pedals
+        if pedal_threshold is not None:
+            CC_SUSTAIN_PEDAL = 64
+            time_pedal_on = 0
+            is_pedal_on = False
+            for cc in [_e for _e in self.control_changes
+                       if _e.number == CC_SUSTAIN_PEDAL]:
+                time_now = int(cc.time*fs)
+                is_current_pedal_on = (cc.value > pedal_threshold)
```
cghawthorne: Very minor nit, >= 64 seems like a better threshold to me because at 64, you're in the upper half of values.

Likely doesn't matter at all because for this kind of sustain processing, we more or less expect it to be just on or off.

pretty_midi/instrument.py
```diff
@@ -113,6 +120,29 @@ def get_piano_roll(self, fs=100, times=None):
...
+                    time_pedal_on = time_now
+                    is_pedal_on = True
+                elif is_pedal_on and not is_current_pedal_on:
+                    # For each pitch, a sustain pedal "retains"
+                    # the maximum velocity up to now due to
+                    # logarithmic nature of human loudness perception
+                    subpr = piano_roll[:, time_pedal_on:time_now]
+
+                    # Take the running maximum
+                    pedaled = np.maximum.accumulate(subpr, axis=1)
```
cghawthorne: I'm not totally sure this is the correct way to handle repeated notes with sustain. The way we did it on magenta was to extend each note within a sustain to either the beginning of the next note of the same pitch or the end of the sustain period, whichever happens first:

https://github.com/tensorflow/magenta/blob/master/magenta/music/sequences_lib.py#L972

The pianoroll equivalent of that would be to just retain whatever the most recent velocity was, not the maximum. My argument for using the most recent velocity would be that if you play a note with high velocity on a piano with sustain on, then wait a long time, then play the same note with low velocity, you won't hear a loud note, you'll hear a soft note.

That said, I could see it going the other way too (play a soft note immediately after a loud note), but I think retaining the most recent velocity better captures what's going on without making too many assumptions about the acoustic properties of the instrument.

What do you think?

maezawa-akira: Hi, I totally understand your point, but I opted to taking the running maximum for three reasons.

1. This approach seemed faster.  When generating piano rolls, this approach seems faster than keeping track of the current note on events, especially when the piano roll w/o pedal messages are already given.  
2. This approach covers the majority of use-case of CC64 on musical instruments. Using the most recent velo is a good assumption for decaying sounds that are played with a long interval.  Considering that pedals are often depressed for only a few bars, and also since CC64 are also used extensively on non-decaying instruments like string pads and organs, I argue that taking the running maximum captures the majority of the use-cases of CC64.    
3. I think the two approaches are complementary ways of approximating the pedals, but the choice would not influence most use-cases of how piano-rolls are used, as binary indicator of note presence.  If knowing the actual loudness is important, then one should convolve the note-on messages with the ADSR envelope.  But this kind of implementation is too time-consuming to be valuable in most cases of interest of using the piano-rolls like AMT or informed source separation; the interest is in knowing the existence of notes and not the loudness.

pretty_midi/instrument.py
maezawa-akira: Thanks for the good point! I fixed the code accordingly.

pretty_midi/instrument.py
cghawthorne: Yup, I can see the arguments that way too. I'll leave it to @craffel to decide. :-)

craffel:
Merged, thank you so much @maezawa-akira !

PR #143, opened 2018-05-17
Correct GitHub repository URL

yaph:
I clicked on the homepage link on https://pypi.org/project/pretty_midi/ and noticed that it doesn't work.


craffel:
Thanks!

PR #146, opened 2018-07-10
QoL enhacement - get_duration method for Note

Ajk4:


craffel:
Cool. Want to add a simple docstring? Something like
```Python
"""Get the duration of the note in seconds."""
```

craffel:
Thanks!

Issue #152, opened 2018-11-22
Compound meters errors?


apmcleod:
get_downbeats() seems to return incorrect values for me when used with compound meters (6/x, 9/x, 12/x).

I think the error comes from pretty_midi.py, line 716 (and also similarly line 720 for final_ts): `beats[start_beat_idx:end_beat_idx:start_ts.numerator]`

This line skips beats based on the numerator value. However, if the meter is compound, the skip should be `start_ts.numerator // 3` I think.

This also led to me uncovering a 2nd bug in utilities.py, the qpm_to_bpm method:

Currently, compound meters are only handled correctly if they have an 8 in the denominator.

6/4, for example, would return a bpm equal to the quarter note, while the compound meter should have the bpm be equal to 3 quarter notes.

In the returns within the `elif denominator in [8, 16, 32]:` block, all returns are treated as if the denominator is 8. Thus, I think every return inside this block should be changed multiplied by an additional factor based on the denominator.

I think a fix would involve removing the special cases for 1, 2, and 4, and just adding them in as [1, 2, 4, 8, 16, 32]. This (with the additional factor) would then become: (note that I am unsure how the else block where I've put !!!!!!!!!! should be handled)

```
if denominator in [1, 2, 4, 8, 16, 32]:
    # simple triple
    if numerator == 3:
        return quarter_note_tempo * denominator / 4
    # compound meter 6/8*n, 9/8*n, 12/8*n...
    elif numerator % 3 == 0:
        return quarter_note_tempo / 3.0 * denominator / 4
    # strongly assume one denominator equals the beat
    else: !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
        return quarter_note_tempo * denominator / 4
else:
    return quarter_note_tempo
```

apmcleod:
Here's an example piece, Bach BWV 864, fugue, from WTC book I, number 19. It is in 9/8.

[bach-0864-fugue.mid.zip](https://github.com/craffel/pretty-midi/files/2606626/bach-0864-fugue.mid.zip)

apmcleod:
Fixed by #153

PR #153, opened 2018-11-22
Compound meter bug fixes

apmcleod:
This fixes issue #152 in 2 ways:

1. It checks for compound meters when returning downbeats.
2. It checks for compound meters in all denominators when converting qpm to bpm.

### DIFF
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -498,7 +498,13 @@ class PrettyMIDI(object):
         return tempi[0]
 
     def get_beats(self, start_time=0.):
-        """Return a list of beat locations, according to MIDI tempo changes.
+        """Returns a list of beat locations, according to MIDI tempo changes.
+        For compound meters (any whose numerator is a multiple of 3 greater
+        than 3), this method returns every third denominator note (for 6/8
+        or 6/16 time, for example, it will return every third 8th note or
+        16th note, respectively). For all other meters, this method returns
+        every denominator note (every quarter note for 3/4 or 4/4 time, for
+        example).
 
         Parameters
         ----------
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -712,12 +712,19 @@ class PrettyMIDI(object):
             end_beat_idx = index(beats, end_ts.time, start_beat_idx)
             # Add beats within this time signature range, skipping beats
             # according to the current time signature
-            downbeats.append(
-                beats[start_beat_idx:end_beat_idx:start_ts.numerator])
+            if start_ts.numerator % 3 == 0 and start_ts.numerator != 3:
+                downbeats.append(beats[
+                    start_beat_idx:end_beat_idx:(start_ts.numerator // 3)])
+            else:
+                downbeats.append(beats[
+                    start_beat_idx:end_beat_idx:start_ts.numerator])
         # Add in beats from the second-to-last to last time signature
         final_ts = time_signatures[-1]
         start_beat_idx = index(beats, final_ts.time, end_beat_idx)
-        downbeats.append(beats[start_beat_idx::final_ts.numerator])
+        if final_ts.numerator % 3 == 0 and final_ts.numerator != 3:
+            downbeats.append(beats[start_beat_idx::(final_ts.numerator // 3)])
+        else:
+            downbeats.append(beats[start_beat_idx::final_ts.numerator])
         # Convert from list to array
         downbeats = np.concatenate(downbeats)
         # Return all downbeats after start_time
--- a/pretty_midi/utilities.py
+++ b/pretty_midi/utilities.py
@@ -237,26 +237,17 @@ def qpm_to_bpm(quarter_note_tempo, numerator, denominator):
             'Time signature denominator must be an int greater than 0, but {} '
             'was supplied.'.format(denominator))
 
-    # denominator is whole note
-    if denominator == 1:
-        return quarter_note_tempo / 4.0
-    # denominator is half note
-    elif denominator == 2:
-        return quarter_note_tempo / 2.0
-    # denominator is quarter note
-    elif denominator == 4:
-        return quarter_note_tempo
-    # denominator is eighth, sixteenth or 32nd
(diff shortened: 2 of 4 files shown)


craffel:
Thanks for this. I honestly don't know enough to know whether this is correct - are there any tests you could write to make sure it does what you intend? Is the behavior already correct for other unusual time signatures (7/12, or something)?

apmcleod:
I added some tests here. Extensive testing for qpm_to_bpm, which works as I'd expect, and one simple test for get_downbeats, since the other functionality is all tested already in the method. This includes testing for unusual numerators.

Irregular denominators are more tricky. For those falling outside of [1,2,4,8,16,32], I've just left the functionality as it was previously (to just return qpm as bpm). I'm actually not even sure that MIDI allows for such time signatures, since its denominators are stored as powers of 2, although there may be some way to fudge this with the other fields in the MIDI spec.  I'm not even sure what the correct value would be in those cases. One option would be to remove the large `if denominator in [1,2,4,8,16,32]:` check and making it the global case for all denominators, but I'm not sure it's really worth it since I don't know how the parser handles irregular denominators in the first place.

apmcleod:
I checked mido, and it will only ever pass through denominators that are powers of 2: https://github.com/mido/mido/blob/master/mido/midifiles/meta.py line 379.

craffel:
Awesome, thank you. This looks good to me but I do lack the background to know if it's 100% correct. @jrgillick is this something you could look at and let me know if it matches your understanding?

jrgillick:
From what I can understand of the code, I think it looks good. 

- Most important to me is that for 6/8, get_beats() should give you every 8th note, and get_downbeats() should give you every 3rd 8th note, like (**1**) 2 3 (**4**) 5 6. Same goes for 9/8, 12/8, etc. So get_downbeats() for 9/8 should get you (**1**) 2 3 (**4**) 5 6 (**7**) 8 9. Is that what your change does?

- 6/4 is where it starts to get hairy. I guess technically it means 2 dotted half notes, so get_downbeats() should give you every 3rd quarter note, also like (**1**) 2 3 (**4**) 5 6. I think that's what your code does? But the convention for 6/4 isn't as clear as for 6/8, so in this case as a user I might like/expect to have get_downbeats() just give me the first beat of every measure, and if I have a more specific meaning of downbeats in mind, I would just use get_beats() and then divide how I see fit.

- This logic extends to more complicated time signatures that could be divided up in multiple ways. For example 7/8 could feel like (**1**) 2 3 (**4**) 5 (**6**) 7 or (**1**) 2 (**3**) 4 (**5**) 6 7. Here i'd expect to have get_downbeats() give me the first beat of the measure and I'd use get_beats() if I want to divide it more. For simplicity, it might be nice for the library not to make any assumptions about downbeats when it's not obvious, and just leave it to the user.

apmcleod:
That isn't quite what the code does at the moment (nor what it used to do). I'll try to summarize:

**Downbeats**
- Previously, for 6/8, 9/8, 12/8, etc., get_beats() returned every 3rd eight note, and get_downbeats() returned the 1st beat of every 3rd measure. This is clearly wrong.

- With this change, get_downbeats() will return the first beat of every measure, regardless of time signature.

-This matches with the definition of a downbeat that I'm familiar with, and how the term is used in MIR literature I'm aware of, so I think this functionality is correct.

**Beats**
- get_beats() is a bit more complicated. For compound meters (6/X, 9/X, 12/X, etc.), the musical definition typically describes the difference between compound and simple meters as: simple meters divide each beat in 2, while compound meters divide each beat in 3. Thus, the beat for 6/8 should be every 3rd eight note (and for 6/4 every 3rd quarter note, etc.). That is what this change does. (Simple meter functionality remains unchanged, returning every quarter note for any 4/4 (or 7/4, or 13/4), for example.)

- The previous version returned every 3rd eight note for 6/8 (and likewise for 6/16 and 6/32), but every quarter note for 6/4, which is inconsistent I think.

- MIR literature is not consistent on beats. Some papers consider 6 beats per 6/8 bar, while some consider 2.

So, this change makes the library consistent with the musical definition of beats, and some of the MIR literature.

To be honest, I'm not sure what you would like the default functionality to be. I would prefer this change's functionality, but I can see the merits either way. It may be useful to add a flag to get=beats() (and/or qpb_to_bpm()???) allowing the user to choose what to do for compound meters. Then we would need to also decide the default functionality.

jrgillick:
Ah, I misunderstood what you were doing - thanks for the explanation and for your good work!

- Downbeats fix looks good
- I'm not totally sold on every 3rd quarter note for 6/4, since though this may be "correct" according to the definitions of simple/compound meter, but I think there's plenty of pop music at least that could be written down in 6/4 but with the beats felt on the quarter and not necessarily in groups of 3 (e.g. https://www.youtube.com/watch?v=MmZexg8sxyk). But as you point out, these terms are not always consistent, so I'm ok with the way you have it. Might be worth adding a comment in the definition of get_beats() about the simple/compound meter definitions.

So @craffel the change looks good to me 👍 

Remaining things that I think could be worth addressing/discussing in the future:

- 7/8, 11/8, 13/8 etc. are not divisible by 2 or 3, so getting quarter notes back from get_beats() is still awkward (though not introduced by this change). Maybe just getting every 8th note in those meters when you call get_beats() would be better?

- I'd find it useful to have a function that would return me the times of every quarter note for X/4 and every eighth note for X/8. This would be convenient for things like quickly synthesizing a metronome. I had initially misunderstood and thought this is part of what you were doing. Maybe a good name for this future function could be get_pulses()?

apmcleod:
Ah, yes, actually that is what this does not in fact! It will return eighth notes for X/8 (assuming not compound) and quarter notes for X/4 (not compound), 16th notes for X/16, half notes for X/2, etc. And in the compound case, it will return every third one of those.

craffel:
Thank you both for the discussion. @apmcleod Do you mind adding this comment to the docstring? 
> Might be worth adding a comment in the definition of get_beats() about the simple/compound meter definitions.

apmcleod:
Done. I wrote a couple of examples there too, so hopefully it is clear. Feel free to edit if you'd like.

craffel:
Looks great! Thanks so much for this.

PR #166, opened 2019-04-11
Add a method for merging subsequent notes

jollyjonson:
This pull request adds a method in the Instrument class for merging subsequent notes in an instrument and a method in the PrettyMIDI object to call this method on all its instruments. All notes that directly follow each other and have the same pitch are merged to one note with a velocity that is the average of all merged notes.

When doing any kind of frame wise pitch predictions I found this method useful to listen to the outcome in a more 'continuos' way than a frame wise cluttered mess... I don't know if this is in general a valuable feature for pretty-midi to have, but I propose it here so you can decide yourself!

### DIFF
--- a/pretty_midi/instrument.py
+++ b/pretty_midi/instrument.py
@@ -339,6 +339,53 @@ class Instrument(object):
         for note in notes_to_delete:
             self.notes.remove(note)
 
+    def merge_subsequent_notes(self):
+        """Merges all notes which are directly following each
+        other and have the same pitch, velocities are averaged.
+
+        """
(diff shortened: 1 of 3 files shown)


craffel:
Thank you for this. I'm not sure this is a common enough operation to go into "core" `pretty_midi`; I think it would make more sense as an example in the examples/ directory. WDYT?

jollyjonson:
Thanks for your reply! Yes, the examples seem to be the more appropriate place for this. I'll change this PR accordingly when I find the time.

PR #176, opened 2020-02-23
added option to clip input files for questionable files

### LINKED ISSUE #164: clip values outside 0-127
mido has an optional `clip` argument to `MidiFile` class, which can be seen [here](https://github.com/mido/mido/pull/110). There should be a corresponding `clip` argument for `PrettyMIDI`.

sleep-yearning:
default behavior doesn't change, but if the clip parameter is passed as True it's passed on to the mido backend which already implements the same behavior.

This should resolve #164 
Edit: might increase the version number for mido requirement

### DIFF
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -48,7 +48,7 @@ class PrettyMIDI(object):
         List of :class:`pretty_midi.Lyric` objects.
     """
 
-    def __init__(self, midi_file=None, resolution=220, initial_tempo=120.):
+    def __init__(self, midi_file=None, resolution=220, initial_tempo=120.,clip=False):
         """Initialize either by populating it with MIDI data from a file or
         from scratch with no data.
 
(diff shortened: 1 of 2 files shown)


sleep-yearning:
After looking through the mido versions, the required version should be 1.2.8, that's when clipping got implemented there.
Idk if that's the reason for the Travis Check fail though.

craffel:
Hey, thanks for this. To be honest I don't think that corrupt MIDI files should be read in this manner - it seems dangerous to arbitrarily clip any non-127 data byte to 127. If a `pretty_midi` user wanted this behavior, I'd suggest they load in the MIDI file with mido with `clip=True`, write it out to a tempfile, and then read it in with `pretty_midi`. I could be convinced otherwise. Either way, you'd need to add the `clip` argument to the docstring.

sleep-yearning:
Alright, I get where you're coming from. An explanation why it was useful for me: 
I wanted to use pretty-midi to analyze and convert large sets of midi files. 
Having files in there which stop the whole process is pretty cumbersome, especially if you can be tolerant for _some_ values being off.

I switched from mido to pretty-midi in my code and had to change it anyway for me to work.
And since another user requested this too, I added the PR. 
Let me know if you want to pull or not, I will add the docstring then.

craffel:
In that case I would suggest wrapping your MIDI file loading in a try/except and skipping the problematic files. If a data byte is > 127 I would treat the file as corrupt, i.e. not useful.

PR #179, opened 2020-04-01
add some condition for recognize drum track

lzxzy:
For some drum tracks, the **Note** duration is very short. So it's the start time and end time could be in a tick, which may cause can't recognize the track successful. So I add the condition for recognizing the drum better.


craffel:
Thanks for the PR, but this would silently change the information in a MIDI file, which would be a surprising and unintended side-effect in some use-cases. It would be much better to manually fix the MIDI files before loading them into `pretty_midi`.

lzxzy:
Thanks for your reply, I'll reference your advice latter.

PR #189, opened 2020-10-13
Added vocals instrument recognition and extended lyrics for text events

gulnazaki:
By counting maximum number of instrument's note-ons and lyrics events coincidences, the instrument that corresponds to vocals is marked with is_vocals. Also, each lyrics object contains the corresponding note, if exists (not whitespace). Finally, load_metadata takes into account all tracks and lyrics are also created for text events (not only lyrics events).

### DIFF
--- a/pretty_midi/containers.py
+++ b/pretty_midi/containers.py
@@ -179,7 +179,7 @@ class KeySignature(object):
 
 
 class Lyric(object):
-    """Timestamped lyric text.
+    """Timestamped lyric text with corresponding note.
 
     Attributes
     ----------
@@ -187,15 +187,21 @@ class Lyric(object):
         The text of the lyric.
     time : float
         The time in seconds of the lyric.
+    note: :class:`pretty_midi.Note` object
+        The note that corresponds to lyric, None for empty text.
     """
 
     def __init__(self, text, time):
         self.text = text
         self.time = time
+        self.note = None
+
+    def set_note(self, note):
+        self.note = note
 
     def __repr__(self):
-        return 'Lyric(text="{}", time={})'.format(
-            self.text.replace('"', r'\"'), self.time)
+        return 'Lyric(text="{}", time={}, note={})'.format(
+            self.text.replace('"', r'\"'), self.time, self.note)
 
     def __str__(self):
-        return '"{}" at {:.2f} seconds'.format(self.text, self.time)
+        return '"{}" at {:.2f} seconds, {}'.format(self.text, self.time, self.note)
--- a/pretty_midi/instrument.py
+++ b/pretty_midi/instrument.py
@@ -34,6 +34,8 @@ class Instrument(object):
         The program number of this instrument.
     is_drum : bool
         Is the instrument a drum instrument (channel 9)?
+    is_vocals : bool
+        Is the 'instrument' a vocalist, synchronized with the lyrics
     name : str
         Name of the instrument.
     notes : list
@@ -51,10 +53,14 @@ class Instrument(object):
         """
         self.program = program
         self.is_drum = is_drum
+        self.is_vocals = False
         self.name = name
         self.notes = []
         self.pitch_bends = []
         self.control_changes = []
+
+    def set_as_vocals(self):
+        self.is_vocals = True
 
     def get_onsets(self):
         """Get all onsets of all notes played by this instrument.
@@ -530,5 +536,5 @@ class Instrument(object):
         return synthesized
 
     def __repr__(self):
-        return 'Instrument(program={}, is_drum={}, name="{}")'.format(
-            self.program, self.is_drum, self.name.replace('"', r'\"'))
+        return 'Instrument(program={}, is_drum={}, is_vocals={}, name="{}")'.format(
+            self.program, self.is_drum, self.is_vocals, self.name.replace('"', r'\"'))
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -118,6 +118,8 @@ class PrettyMIDI(object):
             self.time_signature_changes = []
             # Empty lyrics list
             self.lyrics = []
+            # Empty lyrics_times dictionary (time,lyrics_index)
+            self.__lyrics_times = {}
 
     def _load_tempo_changes(self, midi_data):
         """Populates ``self._tick_scales`` with tuples of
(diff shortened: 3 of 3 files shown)


pretty_midi/containers.py
```diff
@@ -179,23 +179,29 @@ def __str__(self):
...
+    """Timestamped lyric text with corresponding note.
 
     Attributes
     ----------
     text : str
         The text of the lyric.
     time : float
         The time in seconds of the lyric.
+    note: :class:`pretty_midi.Note` object
+        The note that corresponds to lyric, None for empty text.
```
craffel, line 191: Not sure what you mean by "empty text" here?

pretty_midi/containers.py
```diff
@@ -179,23 +179,29 @@ def __str__(self):
...
+        self.note = None
+
+    def set_note(self, note):
+        self.note = note
 
     def __repr__(self):
-        return 'Lyric(text="{}", time={})'.format(
-            self.text.replace('"', r'\"'), self.time)
+        return 'Lyric(text="{}", time={}, note={})'.format(
+            self.text.replace('"', r'\"'), self.time, self.note)
```
craffel, line 204: If you use self.note in this format string, it will use self.note.__str__() which will make the __repr__ of Lyric invalid. You should use self.note.__repr__() here instead.

pretty_midi/instrument.py
```diff
@@ -34,6 +34,8 @@ class Instrument(object):
         The program number of this instrument.
     is_drum : bool
         Is the instrument a drum instrument (channel 9)?
+    is_vocals : bool
+        Is the 'instrument' a vocalist, synchronized with the lyrics
```
craffel, line 38: It wouldn't be synchronized with the lyrics if self.is_vocals is set manually; I would just say "Is the instrument a vocalist?"

pretty_midi/instrument.py
```diff
@@ -51,11 +53,15 @@ def __init__(self, program, is_drum=False, name=''):
         """
         self.program = program
         self.is_drum = is_drum
+        self.is_vocals = False
         self.name = name
         self.notes = []
         self.pitch_bends = []
         self.control_changes = []
 
+    def set_as_vocals(self):
```
craffel, line 62: The convention in pretty_midi is to not include getter/setter functions and to instead allow public manipulation of instance attributes (e.g. the user should just do Instrument.is_vocals = True instead of calling Instrument.set_as_vocals()). So, I would delete this method.

pretty_midi/pretty_midi.py
```diff
@@ -166,27 +168,34 @@ def _load_metadata(self, midi_data):
...
-            elif event.type == 'time_signature':
-                ts_obj = TimeSignature(event.numerator,
-                                       event.denominator,
-                                       self.__tick_to_time[event.time])
-                self.time_signature_changes.append(ts_obj)
-
-            elif event.type == 'lyrics':
-                self.lyrics.append(Lyric(
-                    event.text, self.__tick_to_time[event.time]))
+        # We search for metadata events on all tracks
```
craffel, line 177: My understanding is that if key signature or time signature events appearing on a track other than 0, the MIDI file should be treated as corrupt. Do you have any reference to the contrary?

pretty_midi/pretty_midi.py
```diff
@@ -166,27 +168,34 @@ def _load_metadata(self, midi_data):
...
+                                           self.__tick_to_time[event.time])
+                    self.time_signature_changes.append(ts_obj)
+
+                # Lyrics at time 0 are related to copyright, author, title, etc. so we skip them
+                elif (event.type == 'lyrics' or event.type == 'text') and self.__tick_to_time[event.time] != 0:
+                    # keep (start_time, lyrics index) except for line breaks, that don't correspond to notes
+                    if not event.text.isspace():
+                        self.__lyrics_times[self.__tick_to_time[event.time]] = len(self.lyrics)
+                    self.lyrics.append(Lyric(
+                        event.text, self.__tick_to_time[event.time]))
```
craffel, line 198: I would like you to keep the lyrics processing the same - pretty_midi is not designed to make opinionated decisions about which lyric events should be considered valid (e.g. not at time zero and not newline). These may be useful to some users even if they are not useful to you.

I'd also prefer to store text in a separate self.text_events list, since they are often used for different purposes than lyrics.

pretty_midi/pretty_midi.py
```diff
@@ -382,6 +397,16 @@ def __get_instrument(program, channel, track, create_new):
                         program, event.channel, track_idx, 0)
                     # Add the control change event
                     instrument.control_changes.append(control_change)
+            # Find instument with maximum lyrics matches
+            channel = np.argmax(lyrics_matches)
```
craffel, line 401: Many MIDI files will not have any lyrics instruments (i.e. lyrics_matches will be zero for all instruments) and some MIDI files will have more than one. Why not consider an instrument a vocal instrument if there is at least one lyric event?

craffel (changes requested):
Hi, thanks a lot for proposing this change. I have some minor changes requested, in addition to some larger structural changes. At a high-level I think it may make more sense to include most this functionality in downstream processing of the PrettyMIDI object. Specifically, I don't think the metadata processing should change in any way, and I'd rather not use heuristics to filter out lyrics and determine whether an instrument is a vocal instrument or not - that can be left for downstream user code. I think the main piece of functionality that this PR implements that is a limitation of pretty_midi is the storage of text events, which I would be happy to have. The rest I think can be handled by user code (e.g. you could create a list of Lyric, Note tuples to associate Lyrics/Text Events with Notes; you could manually determine which instrument is a vocal instrument based on whatever heuristics you want, etc).

pretty_midi/containers.py
gulnazaki: I meant whitespace (such as '\r'), used for lyrics formatting.

pretty_midi/containers.py
gulnazaki: You are right, noted

pretty_midi/instrument.py
gulnazaki: Wrong choice of words, I was referring to the procedure, noted.

pretty_midi/pretty_midi.py
gulnazaki: In some files text events appear on other channels (e.g. 69df087321438541ae7c8f064b109917.mid).

pretty_midi/pretty_midi.py
gulnazaki: I was thinking of classifying both lyrics and text events as:
- **text** if they appear at time 0
- **lyrics** if they appear at any other time

since most of the time midi annotators don't distinguish between text and lyrics. What is your opinion on this?

pretty_midi/pretty_midi.py
gulnazaki: It is not that simple, since a lyric event may coincide with multiple instruments, i.e. a guitar strum and a vocal note simultaneously. This is the method MuseScore uses to discover vocal instruments, for example. It could fail in the event of multiple singers, obviously.

gulnazaki:
I think it would be better to close this PR, if you believe the vocals instrument funcionality should be better left to user's code. I can open a new one, where I add the text events as an attribute. I can do it the straightforward way, or I think it would more insightful to classify lyrics and text based on time (as I mention [above](https://github.com/craffel/pretty-midi/pull/189#discussion_r504825539)). Also, a `load_lyrics_and_text` function could deal with lyrics and text alone, on all tracks. Let me know what you think.

PR #195, opened 2020-11-06
Fix get_piano_roll pedal implementation  

TimFelixBeyer:
I'm not entirely sure whether I precisely understand the desired output of get_piano_roll. 
However, after comparing it with https://github.com/magenta/note-seq/tree/master and the piano-roll from there I think there are two small issues with it:
One is an off-by-one error and the other happens when a note offset happens just before/after the pedal is pressed down:

Due to rounding, the pedal might be pressed after the note offset has already occurred.
For example, both offset a and offset b round to t0, and would be held:
t0-->t0+fs
|      *     |  pedal
| *          | offset a -> should not be held
|          * | offset b -> should be held

Again I'm not sure whether the spec of get_piano_roll is different from other piano rolls. 
I think regardless of that, the pedal issue should be addressed.

### DIFF
--- a/pretty_midi/instrument.py
+++ b/pretty_midi/instrument.py
@@ -124,18 +124,31 @@ class Instrument(object):
         if pedal_threshold is not None:
             CC_SUSTAIN_PEDAL = 64
             time_pedal_on = 0
+            time_pedal_on_exact = 0
             is_pedal_on = False
+            # Keep track of non-quantized offset times
+            end_times = np.zeros((128, int(fs * end_time)))
+            for note in self.notes:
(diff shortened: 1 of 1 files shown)


craffel:
@maezawa-akira added the sustain pedal functionality, maybe he can comment?

TimFelixBeyer:
It seems like @maezawa-akira is not active on GitHub anymore.
How would you like to proceed?

page200:
> after comparing it with https://github.com/magenta/note-seq/tree/master and the piano-roll from there

@TimFelixBeyer I can't find the documentation under this link. Do you have code at hand for creating a piano roll from a MIDI file in `magenta`?

PR #201, opened 2021-05-03
User-defined file loading

idoby:
Allow passing in a loaded mido object to allow more fine-grained control over how mido loads the MIDI file.

### DIFF
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -57,8 +57,11 @@ class PrettyMIDI(object):
 
         """
         if midi_file is not None:
+            # File passed in is already loaded, just use it
+            if isinstance(midi_file, mido.MidiFile):
+                midi_data = midi_file
             # Load in the MIDI data using the midi module
-            if isinstance(midi_file, six.string_types):
+            elif isinstance(midi_file, six.string_types):
                 # If a string was given, pass it as the string filename
                 midi_data = mido.MidiFile(filename=midi_file)
             else:


a-pillay:
I support this PR (assuming conflicts can be resolved in a straightforward way)! This is an extremely convenient way to implement stuff like changing the midi file tempo which appears to be quite complicated to do within the pretty-midi domain.

craffel:
Not sure how I missed this but thanks for flagging it @a-pillay . My only suggestion is just to provide a new argument rather than overloading midi_file since this certainly isn't a "midi file" when it's a mido object.

idoby:
Then you run into the trouble of having to check that both arguments weren't passed in.
I think this kind of interface is less error-prone for the user as well.
Either way, I'm not sure I have the time to fix this...

craffel:
I wouldn't call that trouble - it's a single if statement, and adding a single if statement is better than making an argument's meaning ambiguous, because the arguments are user-facing.

craffel:
@a-pillay feel free to pick up this PR if you have time.

a-pillay:
Sure, I can take a look at it. 

To summarize, this change would involve:
1. Creating a new argument `mido_object` in `pretty_midi.PrettyMIDI` with default=_None_
2. Handling passed in `mido.MidiFile` via a new if-else block:
3. This block would be the first one for priority (alternately should we flag situations when both `mido_object` and `midi_file` arguments are populated?)

Could you please confirm this logic before we implement the same?

craffel:
Yes, that would be fine. I don't think the order of the blocks is terribly important as long as all cases are handled.

a-pillay:
I have created  #241 to implement this feature.

PR #215, opened 2022-10-21
Syntax fix for equality check with 0.

kevinzakka:
Hi @craffel,

Big fan of your research -- was pleasantly surprised to see you are the author of this library. Been very fun using it over the last few weeks.

This is just a small fix for python 3.8+.

Cheers


craffel:
Ah wonderful, what a throwback to a time when I didn't realize that this was wrong. Thanks.

PR #216, opened 2022-12-19
updated setup.py with option to install tests

George-Ogden:
now installs `pytest` with 
```
pip install pretty_midi[test]
```
to make development easier
> was hoping to make other modifications but that might be a future project


craffel:
Thanks, any reason you think 0.6.0 is needed?

George-Ogden:
There isn't an important reason for this - most people will end up installing the most recent version of pytest. I include it because I have done on previous projects.

PR #225, opened 2023-05-01
Allow pathlib.Path as input for PrettyMIDI init

almostimplemented:
Minor convenience PR to accept `pathlib.Path` when passing a filename as input to `pretty_midi.PrettyMIDI`.


craffel:
I'm out of touch with best practices - is this better than just requiring the user to pass `str(path)` on their local `Path` instance (`path` in my example)? Like, is it best practice to just be passing around and manipulating `Path`s and never referring to paths as strings?

almostimplemented:
I think typically, when a client passes a filename, they expect both string and `pathlib.Path` types to be valid input.

I can't speak with authority on best practice, but for example, `numpy.load/save`, `torch.load`, and `mido.MidiFile` all permit strings and `Path` objects.

It looks like `torch` has a type annotation of `FILE_LIKE: TypeAlias = Union[str, os.PathLike, BinaryIO, IO[bytes]]`, whereas `numpy` checks `if hasattr(file, 'read') # or: , 'write')`. When the argument is explicitly a filename (as in the case of `mido`), you probably would not do this check and instead simply pass the value into `open`.

Issue #220, opened 2023-04-04
Type Mismatch in Calling `fluidsynth.py` 


Yao-Lirong:
In `fluidsynth` of `instrument.py` (attached as the bottom last cell), there is this following function call to `fluidsynth.py` (attached as the bottom second cell). The last line calls `sfload`, which should take in an argument of `c_char_p` type. However, `sf2_path`, according to spec of this function, is a python `str`. This causes the error: 

```python
return fluid_synth_sfload(self.synth, filename, update_midi_preset)
ArgumentError: argument 2: <class 'TypeError'>: wrong type
```

A simple fix would be to add the following line before calling `sfload`: (A pull request will be opened soon to provide this simple fix) 

```python
        sf2_path = c_char_p(sf2_path.encode())
        sfid = fl.sfload(sf2_path)
```

The following is the definition of `sfload` in `fluidsynth.py`:

```python
    fluid_synth_sfload = cfunc('fluid_synth_sfload', c_int,
                           ('synth', c_void_p, 1),
                           ('filename', c_char_p, 1),
                           ('update_midi_presets', c_int, 1))
    def sfload(self, filename, update_midi_preset=0):
        """Load SoundFont and return its ID"""
        return fluid_synth_sfload(self.synth, filename, update_midi_preset)
```

The following is the original definition of `fluidsynth` in `instrument.py`. 

```python

    def fluidsynth(self, fs=44100, sf2_path=None):
        """Synthesize using fluidsynth.

        Parameters
        ----------
        fs : int
            Sampling rate to synthesize.
        sf2_path : str
            Path to a .sf2 file.
            Default ``None``, which uses the TimGM6mb.sf2 file included with
            ``pretty_midi``.

        Returns
        -------
        synthesized : np.ndarray
            Waveform of the MIDI data, synthesized at ``fs``.

        """
        # If sf2_path is None, use the included TimGM6mb.sf2 path
        if sf2_path is None:
            sf2_path = pkg_resources.resource_filename(__name__, DEFAULT_SF2)

        if not _HAS_FLUIDSYNTH:
            raise ImportError("fluidsynth() was called but pyfluidsynth "
                              "is not installed.")

        if not os.path.exists(sf2_path):
            raise ValueError("No soundfont file found at the supplied path "
                             "{}".format(sf2_path))

        # If the instrument has no notes, return an empty array
        if len(self.notes) == 0:
            return np.array([])

        # Create fluidsynth instance
[...]

PR #226, opened 2023-05-31
Revert "fix type mismatch in `fluidsynth` calling `sfload`"

DavidDiazGuerra:
Reverts craffel/pretty-midi#221

That PR crashes with the newer versions of pyfluidsynth, it is fixing a bug that was already fixed in pyfluidsynth in 2017. 

See https://github.com/craffel/pretty-midi/issues/220#issuecomment-1569634118

### DIFF
--- a/pretty_midi/instrument.py
+++ b/pretty_midi/instrument.py
@@ -9,7 +9,6 @@ except ImportError:
     _HAS_FLUIDSYNTH = False
 import os
 import pkg_resources
-from ctypes import c_char_p
 
 from .containers import PitchBend
 from .utilities import pitch_bend_to_semitones, note_number_to_hz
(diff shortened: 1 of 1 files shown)


DavidDiazGuerra:
Agree! I'll try to find out which one is the oldest `pyfluidsynth` release that works with this PR and commit an updated `setup.py` with it.

DavidDiazGuerra:
I can confirm that this PR works well with any `pyfluidsynth>=1.3.1`. I really think it should also work with any `pyfluidsynth>=1.2.5`, but I couldn't test this since those versions seem to not be compatible with my FluidSynth 3 installation.

Should I include `pyfluidsynth>=1.3.1` as an `extras_require` in `setup.py` or should we wait till someone can test it on `pyfluidsynth==1.2.5`?

craffel:
I think 1.3.1 is fine, thanks.

craffel:
Thanks!

PR #230, opened 2023-07-14
Pass sample rate to Instrument.fluidsynth

jvlmdr:
The sample-rate `fs` is only passed implicitly with the synthesizer. However, it is used within Instrument.fluidsynth to compute note positions at the given sample rate.

### DIFF
--- a/pretty_midi/fluidsynth.py
+++ b/pretty_midi/fluidsynth.py
@@ -31,7 +31,7 @@ def get_fluidsynth_instance(synthesizer=None, sfid=0, fs=None):
     fs : int
         Sampling rate to synthesize at.
         Default ``None``, which falls back to ``pretty_midi.fluidsynth.DEFAULT_SAMPLE_RATE``
-        if a new instance must be created.
+        = 44100 if a new instance must be created.
         If ``synthesizer`` is an existing instance and ``fs`` is specified, then
         ValueError will be raised if the sample rates are not equal.
 
--- a/pretty_midi/fluidsynth.py
+++ b/pretty_midi/fluidsynth.py
@@ -57,6 +57,10 @@ def get_fluidsynth_instance(synthesizer=None, sfid=0, fs=44100):
         sfid = synthesizer.sfload(sf2_path)
         new_instance_created = True
     elif isinstance(synthesizer, fluidsynth.Synth):
+        if fs:
+            synth_fs = synthesizer.get_setting('synth.sample-rate')
+            if synth_fs != fs:
+                raise ValueError("synth sample-rate does not match fs", synth_fs)
         new_instance_created = False
     else:
         raise ValueError("synthesizer must be a str or a fluidsynth.Synth instance")
--- a/pretty_midi/fluidsynth.py
+++ b/pretty_midi/fluidsynth.py
@@ -12,9 +12,10 @@ except ImportError:
     _HAS_FLUIDSYNTH = False
 
 DEFAULT_SF2 = 'TimGM6mb.sf2'
+DEFAULT_SAMPLE_RATE = 44100
 
 
-def get_fluidsynth_instance(synthesizer=None, sfid=0, fs=44100):
+def get_fluidsynth_instance(synthesizer=None, sfid=0, fs=None):
     """ Check if a valid fluidsynth.Synth instance is provided, and if not,
     create one.
 
@@ -29,8 +30,9 @@ def get_fluidsynth_instance(synthesizer=None, sfid=0, fs=44100):
         Default ``0``, which uses the first soundfont.
     fs : int
         Sampling rate to synthesize at.
-        Only used when a new instance of fluidsynth.Synth is created.
-        Default ``44100``.
+        Default ``None``, which falls back to 44100 if a new instance must be created.
+        If ``synthesizer`` is an existing instance and ``fs`` is specified, then
+        a check is performed to ensure that its sample rate is equal to ``fs``.
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


craffel:
The trouble is that if a pre-created `fluidsynth.Synth` is passed via the `synthesizer` argument, `fs` won't necessarily match the sampling rate of that synthesizer. Is there a way to retrieve the sampling rate of `synthesizer` directly? @DavidDiazGuerra

DavidDiazGuerra:
I didn't realize that `Instrument.fluidsynth` was using `fs` for more stuff than creating the fluidsynth instance. This is indeed a bug and will create problems when using a sampling frequency different than 44,1 kHz. Sorry about this.

The sampling frequency of a fluidsynth instance can be retrieved by `synthesizer.get_setting('synth.sample-rate')`. I think  the best solution for this would be leaving `PrettyMIDI.fluidsynth` as it is in `pretty-midi/main` now and modifying `Instrument.fluidsynth` so, when a fluidsynth instance is provided, its sampling frequency is retrieved with that function instead of using the argument function.

I'm thinking that maybe it could be interesting raising a warning (or even an error) from `pretty_midi.fluidsynth.get_fluidsynth_instance` if both `fs` and a fluidsynth instance are provided, at least if `fs` is different than the sampling rate of the fluidsynth instance.

jvlmdr:
I modified the code to raise a ValueError if both `fs` and `synthesizer` are passed but the sample rates don't match.

(The other option would be to create a new synthesizer if they don't match.)

DavidDiazGuerra:
I think that's the best option since, in order to create a new instance of fluidsynth, we would need the path to the sfz file.

Now I'm wondering what would be the best way to deal with the default values of `fs` in `PrettyMIDI.fluidsynth` and `Instrument.fluidsynth`. In the current implementation, not indicating the value of `fs` when using an instance of fluidsynth with a sampling rate different than 44.1 kHz (the default value) will raise an error when it was quite clear that the user just wanted to continue using the same sampling rate of the fluidsynth instance. Maybe the best option would be leaving the default value to `None` and then managing this inside of `pretty_midi.fluidsynth.get_fluidsynth_instance` and returning the sampling rate of the synthesizer if it was `None`? But I'm not sure, maybe we're adding too much functionality to `pretty_midi.fluidsynth.get_fluidsynth_instance`. What do you think @craffel?

DavidDiazGuerra:
I hadn't seen the last commit when I wrote my previous message. That's indeed in the line of what I was saying, but I'm not sure how it would work since I think `not fs` would be `True` only if `fs == 0`.

jvlmdr:
I just modified `Instrument.fluidsynth` to take `fs` from `synthesizer` if falsey. I think you're right, though, that the default kwarg of 44100 should be removed to allow `fluidsynth(synthesizer=synthesizer)`. Maybe we should change it to None (and we could use 44100 if `fs is None and synthesizer is None` to avoid breaking backwards compatibility).

jvlmdr:
Cool! Maybe `fs is None` rather than `not fs` would be more explicit

pretty_midi/fluidsynth.py
```diff
@@ -53,10 +55,14 @@ def get_fluidsynth_instance(synthesizer=None, sfid=0, fs=44100):
...
         if not os.path.exists(synthesizer):
             raise ValueError("No soundfont file found at the supplied path {}".format(sf2_path))
+        fs = fs or DEFAULT_SAMPLE_RATE
         synthesizer = fluidsynth.Synth(samplerate=fs)
         sfid = synthesizer.sfload(sf2_path)
         new_instance_created = True
     elif isinstance(synthesizer, fluidsynth.Synth):
+        synth_fs = synthesizer.get_setting('synth.sample-rate')
+        if fs and synth_fs != fs:
+            raise ValueError("synth sample-rate does not match fs", synth_fs)
```
craffel, line 65: May as well make the error message more helpful `f"synthesizer sample rate of {synth_fs} doesn't match provided fs of {fs}"`

pretty_midi/instrument.py
```diff
@@ -423,15 +423,15 @@ def synthesize(self, fs=44100, wave=np.sin):
...
         """Synthesize using fluidsynth.
 
         Parameters
         ----------
         fs : int
             Sampling rate to synthesize at.
-            Only used when a new instance of fluidsynth.Synth is created.
-            Default ``44100``.
+            Default ``None``, which takes the sampling rate from ``synthesizer``, or
+            uses 44100 if ``synthesizer`` has not been created.
```
craffel, line 434: or uses pretty_midi.fluidsynth.DEFAULT_SAMPLE_RATE = 44100 if synthesizer...

pretty_midi/instrument.py
```diff
@@ -465,7 +465,10 @@ def fluidsynth(self, fs=44100, synthesizer=None, sfid=0, sf2_path=None):
             return np.array([])
 
         # Create a fluidsynth instance if one wasn't provided
+        # Raises if sample rates do not match
```
craffel, line 468: Raises ValueError

pretty_midi/instrument.py
```diff
@@ -465,7 +465,10 @@ def fluidsynth(self, fs=44100, synthesizer=None, sfid=0, sf2_path=None):
...
         synthesizer, sfid, delete_synthesizer = get_fluidsynth_instance(synthesizer, sfid, fs)
+        # Sample rate required below to compute note positions
+        fs = fs or synthesizer.get_setting('synth.sample-rate')
```
craffel, line 471: I think you can just use synthesizer.get_setting('synth.sample-rate') in either case here, right? Because if the synthesizer is created, it will have the sample rate provided to get_fluidsynth_instance

pretty_midi/pretty_midi.py
```diff
@@ -960,15 +960,15 @@ def synthesize(self, fs=44100, wave=np.sin):
...
         """Synthesize using fluidsynth.
 
         Parameters
         ----------
         fs : int
             Sampling rate to synthesize at.
-            Only used when a new instance of fluidsynth.Synth is created.
-            Default ``44100``.
+            Default ``None``, which takes the sampling rate from ``synthesizer``, or
+            uses 44100 if ``synthesizer`` has not been created.
```
craffel, line 971: or uses pretty_midi.fluidsynth.DEFAULT_SAMPLE_RATE = 44100 if synthesizer...

craffel (changes requested):

craffel:
This looks right to me, requesting mostly cosmetic changes. Thanks.

jvlmdr:
Thanks! I think I have addressed all requested changes

craffel:
Thanks!

PR #232, opened 2023-07-26
ValueError is now raised if end < start

### LINKED ISSUE #231: Note starting point can be greater than its ending point
When instantiating a ``Note`` object, it's possible to pass to the ``start`` parameter a value greater than the one passed to ``end``. If this isn't noticed, it may lead to an error which at first is difficult to track down. Apparently, this is caused by the slice ``synthesized[current_sample:end]``, which results in an empty array due to the aforementioned possibility.

After making sure ``start`` was always lower than ``end`` before passing them to the ``Note`` constructor, the problem dissapeared.

```python
/usr/local/lib/python3.10/dist-packages/pretty_midi/pretty_midi.py in fluidsynth(self, fs, sf2_path)
    972             return np.array([])
    973         # Get synthesized waveform for each instrument
--> 974         waveforms = [i.fluidsynth(fs=fs,
    975                                   sf2_path=sf2_path) for i in self.instruments]
    976         # Allocate output waveform, with #sample = max length of all waveforms

/usr/local/lib/python3.10/dist-packages/pretty_midi/pretty_midi.py in <listcomp>(.0)
    972             return np.array([])
    973         # Get synthesized waveform for each instrument
--> 974         waveforms = [i.fluidsynth(fs=fs,
    975                                   sf2_path=sf2_path) for i in self.instruments]
    976         # Allocate output waveform, with #sample = max length of all waveforms

/usr/local/lib/python3.10/dist-packages/pretty_midi/instrument.py in fluidsynth(self, fs, sf2_path)
    522             end = int(fs*(current_time + event[0]))
    523             samples = fl.get_samples(end - current_sample)[::2]
--> 524             synthesized[current_sample:end] += samples
    525             # Increment the current sample
    526             current_time += event[0]

ValueError: operands could not be broadcast together with shapes (0,) (818,) (0,)
```

TomRolB:
Closes #231

### DIFF
--- a/pretty_midi/containers.py
+++ b/pretty_midi/containers.py
@@ -23,6 +23,9 @@ class Note(object):
     """
 
     def __init__(self, velocity, pitch, start, end):
+        if end < start:
+            raise ValueError("end paramater must be greater than start")
+        
         self.velocity = velocity
         self.pitch = pitch
         self.start = start
--- a/pretty_midi/containers.py
+++ b/pretty_midi/containers.py
@@ -24,7 +24,7 @@ class Note(object):
 
     def __init__(self, velocity, pitch, start, end):
         if end < start:
-            raise ValueError("end paramater must be greater than start")
+            raise ValueError("Note end time must be greater than start time")
         
         self.velocity = velocity
         self.pitch = pitch


pretty_midi/containers.py
```diff
@@ -23,6 +23,9 @@ class Note(object):
     """
 
     def __init__(self, velocity, pitch, start, end):
+        if end < start:
+            raise ValueError("end paramater must be greater than start")
```
craffel, line 27: Suggest "Note end time must be greater than start time" as a bit more informative

craffel (changes requested):

craffel (approved):

craffel:
Thanks!

PR #233, opened 2023-08-17
Let the user choose the MIDI charset

Saisimon:
Use `pretty_midi` to add Chinese lyrics to MIDI, and an error will be reported when calling the `write` method. The default charset cannot handle Chinese, so allow users to choose the correct charset when creating PrettyMIDI object.

### DIFF
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -42,6 +42,8 @@ class PrettyMIDI(object):
         Resolution of the MIDI data, when no file is provided.
     initial_tempo : float
         Initial tempo for the MIDI data, when no file is provided.
+    charset : str
+        Charset of the MIDI.
 
     Attributes
     ----------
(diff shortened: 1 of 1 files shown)


pretty_midi/pretty_midi.py
```diff
@@ -66,10 +68,10 @@ def __init__(self, midi_file=None, resolution=220, initial_tempo=120.):
             # Load in the MIDI data using the midi module
             if isinstance(midi_file, six.string_types) or isinstance(midi_file, pathlib.PurePath):
                 # If a string or path was given, pass it as the filename
-                midi_data = mido.MidiFile(filename=midi_file)
+                midi_data = mido.MidiFile(filename=midi_file, charset=charset)
```
craffel, line 71: Is there a way to get mido to detect the charset of an existing file automatically, or at least get it to report the charset, or something other than having to separately specify it?

craffel (commented):
Thanks - left one question

pretty_midi/pretty_midi.py
Saisimon: I'm not familiar with mido. I just referred to the way to handle charsets in the mido [documentation](https://mido.readthedocs.io/en/stable/api.html?highlight=charset#mido.MidiFile). The default charset is latin1. When the user wants to specify a character set, just pass the corresponding charset parameter.

craffel (approved):

Issue #237, opened 2023-11-04
Docs apperance


michalskibinski109:
Please consider setting theme in `conf.py` to 

```python
html_theme = "sphinx_rtd_theme"
```

It is much more functionall.

craffel:
PR welcome!

michalskibinski109:
Ok I did

craffel:
Thanks!

PR #238, opened 2023-11-06
Splits documentation into chapters, adds readme as first page of docs

michalskibinski109:
#237 

1. Change theme 
2.  Splits documentation for every module into different chapter
3. Adds readme as main page to make docs more readable for new user

This is how it looks now:
image
image


craffel:
Thanks.

Issue #214, opened 2022-08-30
_load_metadata does not load lyrics correctly


holgerkirchhoff:
`_load_metadata` seems to assume that _lyrics_ meta events -- similar to tempo and time signature -- are always found on track 0. Looking at the [Standard MIDI Files Specification](https://midi.org/specifications/file-format-specifications/standard-midi-files) this is not required, and I have MIDI files in which the lyrics events are part of the vocal track, which makes sense to me.

In effect, when loading and saving a MIDI file in the current version (0.2.9), all lyrics events of the vocal track are lost. I assume the same applies to key signature changes that are not on track 0.

craffel:
I'm not sure where I read that all meta events (including lyrics events) should be on track 0. Feel free to make a PR.

holgerkirchhoff:
I wonder if lyrics would need to be handled a bit differently in the class? You are currently keeping it as a single member. However, I guess you could have multiple vocal tracks in a MIDI file with different lyrics. Hence it needs to be part of the `Instrument` class.

holgerkirchhoff:
I guess it is quite common when you have a lead singer and backing vocals.
Having lyrics in the right vocal track also makes it easier to assign syllables to corresponding notes. If all lyrics were on track 0, it would not be immediately clear what notes the lyrics belong to.

I'm not sure if I know enough details about the code to make these changes.

craffel:
That makes sense. I won't be able to make these changes myself but if someone makes a PR I can review.

PR #240, opened 2024-01-26
Store lyrics per track

migperfer:
This pull request solves #214. 

Now each instrument has text and lyrics list attributes, that can be populated with MetaMessages for lyrics or text annotations

### DIFF
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -128,10 +128,10 @@ class PrettyMIDI(object):
             self.key_signature_changes = []
             # Empty time signatures changes list
             self.time_signature_changes = []
-            # Empty lyrics list
-            self.lyrics = []
-            # Empty text events list
-            self.text_events = []
+            # Empty lyrics dict
+            self.lyrics = {}
+            # Empty text events dict
+            self.text_events = {}
             # MIDI Charset
             self._charset = charset
 
@@ -186,8 +186,8 @@ class PrettyMIDI(object):
         # signature changes, and lyrics
         self.key_signature_changes = []
         self.time_signature_changes = []
-        self.lyrics = []
-        self.text_events = []
+        self.lyrics = {}
+        self.text_events = {}
 
         for event in midi_data.tracks[0]:
             if event.type == 'key_signature':
@@ -204,8 +204,6 @@ class PrettyMIDI(object):
 
         # We search for lyrics and text events on all tracks
         # Lists of lyrics and text events lists, for every track
-        tracks_with_lyrics = []
-        tracks_with_text_events = []
         for track in midi_data.tracks:
             # Track specific lists that get appended if not empty
             lyrics = []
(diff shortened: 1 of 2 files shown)


pretty_midi/pretty_midi.py
```diff
@@ -219,14 +217,11 @@ def _load_metadata(self, midi_data):
                         event.text, self.__tick_to_time[event.time]))
                     
             if lyrics:
-                tracks_with_lyrics.append(lyrics)
+                self.lyrics[track.name] = lyrics
```
craffel, line 220: Are track names guaranteed to be unique? OTOH I don't think they are so I'm not sure we should key on them.

migperfer: I agree. I changed it to key on the index of the midi tracks, in consonance to how is done in other parts of the code.

pretty_midi/pretty_midi.py
```diff
@@ -186,8 +186,8 @@ def _load_metadata(self, midi_data):
         # signature changes, and lyrics
         self.key_signature_changes = []
         self.time_signature_changes = []
-        self.lyrics = []
-        self.text_events = []
+        self.lyrics = {}
```
craffel, line 189: We don't want to keep parallel lists of lyrics and tests events in both the PrettyMIDI objects and the Instrument objects it contains. In other words, the PrettyMIDI object should not have a lyrics or text_events attribute. These should only be populated when loading from the MIDI file, and then they should be passed on to the Instrument, and the PrettyMIDI object should forget about them. In any PrettyMIDI class method that accesses lyrics or text events, the lyrics and text events of the constitutuent Instruments should be grabbed instead. The logic for populating the lyrics should then go in the __load_instruments method.

pretty_midi/pretty_midi.py
```diff
@@ -461,7 +463,7 @@ def get_end_time(self):
         """
         # Get end times from all instruments, and times of all meta-events
         meta_events = [self.time_signature_changes, self.key_signature_changes,
-                       self.lyrics, self.text_events]
+                       *self.lyrics.values(), *self.text_events.values()]
```
craffel, line 466: Instrument.get_end_time should be modified to take into account lyrics and text events and lyrics/text events should be removed from this list.

pretty_midi/pretty_midi.py
```diff
@@ -1202,13 +1204,18 @@ def adjust_meta(events):
...
 
         # Adjust key signature change event times
         adjust_meta(self.key_signature_changes)
         # Adjust lyrics
-        adjust_meta(self.lyrics)
-        # Adjust text events
-        adjust_meta(self.text_events)
+        for lyrics_key in self.lyrics:
+            adjusted_lyrics = adjust_meta(self.lyrics[lyrics_key])
+            self.lyrics[lyrics_key] = adjusted_lyrics
```
craffel, line 1214: This should be done to each Instrument's lyrics, not self.lyrics

PR #241, opened 2024-02-14
Allow pre-loaded mido.MidiFile objects to be passed as an argument to PrettyMIDI 

### LINKED ISSUE #200: Init with mido object parameter
Hi,
I started working with this library and I noticed it uses mido. 
In my project I have mido objects and I would like to use pretty_midi functions on them.
Instead of saving it to temporary midi file and loading it again, possibility to pass the mido object to PrettyMidi would speed up things.
Or is there any other way I'm not aware of? 

My intention is to select just one track from midi file and generate chroma from it.

a-pillay:
- Implemented `mido_object` argument in PrettyMIDI to pass pre-loaded `mido.MidiFile` objects
- Updated docstring with the aforementioned change
- Implemented a unit test to verify PrettyMIDI object initialization logic

This closes #200 and #201

### DIFF
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -34,10 +34,14 @@ class PrettyMIDI(object):
 
     Parameters
     ----------
+    mido_object : mido.MidiFile
+        Pre-loaded `mido.MidiFile` object.
+        Default ``None`` would check if ``midi_file`` is populated instead.
     midi_file : str or file
         Path or file pointer to a MIDI file.
         Default ``None`` which means create an empty class with the supplied
         values for resolution and initial tempo.
+        Note: If ``mido_object`` is not ``None``, this argument is ignored.
     resolution : int
         Resolution of the MIDI data, when no file is provided.
     initial_tempo : float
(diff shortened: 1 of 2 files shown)


craffel:
We actually need to put the new arg at the end of the args list to avoid breaking code that uses positional arguments.

a-pillay:
@craffel I agree.

I have modified `mido_object` to be the last argument in PrettyMIDI. Please let me know if anything else needs to be addressed!

craffel:
Need to change argument order in docstring 😅

a-pillay:
I was considering setting the `mido_object` description as the last one in the argument docstring, but proceeded with keeping it immediately after `midi_file` to highlight how the former will be prioritized.

However, I can set the `mido_object` description as the last one (in sync with the argument order) if you could confirm that it sounds good. I find both options equally logical.

craffel:
Docstring argument lists must match the order of the arguments.

a-pillay:
Sounds good! I have updated the docstrings as suggested. Thanks for putting up with my repeated requests for clarifications!

craffel:
Can you raise a ValueError if both a filename and a mido object are provided? This should not work silently because the correct behavior is ambiguous. The logic would look like
```
if mido_object is not None and midi_file is not None:
   raise ValueError("A midi_file and midi_object cannot both be provided.")
```
and then the rest of the logic can proceed assuming that only one of them has been provided (e.g. you don't need an elif).

a-pillay:
Done! Now, a ValueError will be raised if both `mido_object` and `midi_file` arguments are provided at the same time. Also, updated the concerned docstring and unit test to reflect the same. I have also added tests for verifying these ValueErrors but since the existing tests don't use a toolkit like `unittest`, I have implemented them in an alternate way.

pretty_midi/pretty_midi.py
```diff
@@ -59,19 +65,30 @@ class PrettyMIDI(object):
...
         """
-        if midi_file is not None:
-            # Load in the MIDI data using the midi module
-            if isinstance(midi_file, six.string_types) or isinstance(midi_file, pathlib.PurePath):
-                # If a string or path was given, pass it as the filename
-                midi_data = mido.MidiFile(filename=midi_file, charset=charset)
-            else:
-                # Otherwise, try passing it in as a file pointer
-                midi_data = mido.MidiFile(file=midi_file, charset=charset)
+        if mido_object is not None or midi_file is not None:
```
craffel, line 73: This conditional is no longer necessary.

craffel:
Thanks!

Issue #158, opened 2019-01-16
Fluidsynth produces harsh noise


davidkant:
Fluidsynth synthesis sometimes produces loud harsh noise. This happens when accidentally synthesizing midi pitches that are beyond the range of the chosen instrument. The result should be silent but, i'm guessing due to the normalization, this gets turned into pretty brutal noise.

The quickest fix would be implement [Issue #151 Normalization by default](https://github.com/craffel/pretty-midi/issues/151) and let the user deal with it. Although pretty-midi could also do a quick amplitude check before normalizing. Either is fine by me. Just trying to save my students' ears! Or really save them from getting upset with me...

Thank so much!
David

craffel:
PRs are welcome to resolve #151; I'd be happy to have normalization be via a kwarg which is set to `False` by default.

Issue #151, opened 2018-10-22
Normalization by default


stefan-balke:
Hey,

currently there is a default normalization when using fluidsynth:
https://github.com/craffel/pretty-midi/blob/master/pretty_midi/pretty_midi.py#L940

Could we make this optional? This way different MIDI velocities are ignored...

Fluidsynth is producing 16-bit integer values...this normalization scales it such that the values lie in [-1, 1]. I would rather say we divide the fluidsynth output by 2^16 plus making the normalization optional.

What do folks think?

craffel:
The normalization is a lazy way of addressing the fact that we may be summing together the synthesis of many instruments, so the output of the `PrettyMIDI.fluidsynth` method (before normalization) is `[-len(self.instruments)*2^15, len(self.instruments)*2^15 - 1]`. I am fine with making the normalization disable-able via a keyword argument but I would still advocate for hard-normalizing to `[-1, 1]` when normalization is enabled.

yawjalik:
Hello there, is this issue still up for grabs?

craffel:
Yep!

PR #243, opened 2024-02-20
Add normalize parameter to synthesize and fluidsynth methods

yawjalik:
Added `normalize` parameter such that when set to `True`, hard-normalizes the synthesized output wave to [-1, 1]. Otherwise, divide the wave by `len(self.instruments) * 2^15` to prevent clipping, and lets the user handle accordingly. It is set to `False` by default. This serves as a starting point for tackling #151 and #158.


yawjalik:
This is probably my first ever contribution. Any comments or advice will be greatly appreciated!

yawjalik:
Got it, I've changed them to `True` by default.

craffel:
Thanks!

PR #245, opened 2024-02-27
fix: Remove `pkg_resources`

BenjaminHalko:
`pkg_resources` is depreciated in Python 3.12. This PR replaces it with `importlib.resources`.

### DIFF
--- a/pretty_midi/fluidsynth.py
+++ b/pretty_midi/fluidsynth.py
@@ -3,7 +3,7 @@
 """
 
 import os
-import pkg_resources
+from importlib import resources
 
 try:
     import fluidsynth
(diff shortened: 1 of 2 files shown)


craffel:
Thanks, but pkg_resources is still used to get the default SF2 path: https://github.com/craffel/pretty-midi/blob/280352a75c53901eb9b0e4f4dffa416ad041fa33/pretty_midi/fluidsynth.py#L52 That would need to be fixed (and we'd need to verify that, after applying this PR, the library could be installed with the default SF2 in place and usable by the fluidsynth code).

roszcz:
> Thanks, but pkg_resources is still used to get the default SF2 path:
> 
> https://github.com/craffel/pretty-midi/blob/280352a75c53901eb9b0e4f4dffa416ad041fa33/pretty_midi/fluidsynth.py#L52
> 
> That would need to be fixed (and we'd need to verify that, after applying this PR, the library could be installed with the default SF2 in place and usable by the fluidsynth code).

What has to happen to move forward with this? I'm happy to help with the tests – are there any specific scenarios that have to be considered? It would be great to have `pretty_midi` working in 3.12 :)

craffel:
Does the code in this PR work for you as-is on 3.12, including synthesizing with fluidsynth?

PR #247, opened 2024-07-11
Add crop method to PrettyMidi.

### LINKED ISSUE #246: Method to crop PrettyMIDI object.
Feature request: a method that crops a PrettyMIDI object. The user provides a start time and end time.

This can be trivially implemented using `PrettyMIDI.adjust_times`. However, the documentation for `adjust_times` isn't clear that events outside the provided time ranges are removed. I was halfway to manually implementing it myself by modifying events (and adjusting ticks) when I saw that the [tutorial](https://github.com/craffel/pretty-midi/blob/main/Tutorial.ipynb) alludes to the cropping capabilities of  `adjust_times`, which can then be understood by reading the method's source code.

One possibility would be to add this information to the docs. But I think `pm.crop(start_time, end_time)` is more clear (and searchable) than `pm.adjust_times([start_time, end_time], [start_time, end_time])`.

I would happily contribute the (trivial) code if this is deemed a useful addition.

b-boechat:
This pull request resolves #246.

### DIFF
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -1354,7 +1354,8 @@ class PrettyMIDI(object):
         self._update_tick_to_time(self._tick_scales[-1][0] + 1)
 
     def crop(self, start_time=0.0, end_time=None):
-        """Crops the MIDI object, keeping only the segment from ``start_time`` to ``end_time``.
+        """Crops the MIDI object, keeping only the segment from ``start_time``
+        to ``end_time``.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


pretty_midi/pretty_midi.py
```diff
@@ -1353,6 +1353,37 @@ def adjust_meta(events):
...
+            Segment start time, in seconds. Defaults to 0.0.
+        end_time : float
+            Segment end time, in seconds. Defaults to self.end_time().
+
+        """
+        midi_end_time = self.get_end_time()
+
+        # Enforce that start_time is non-negative and end_time is lower than the MIDI object's end time. 
+        if start_time < 0.0:
+            warnings.warn('start_time is less than 0; automatically adjusting to 0.')
```
craffel, line 1371: This should be a ValueError, and let the user fix it.

pretty_midi/pretty_midi.py
```diff
@@ -1353,6 +1353,37 @@ def adjust_meta(events):
...
+            start_time = 0.0 
+        if end_time is None:
+            end_time = midi_end_time
+        elif end_time > midi_end_time:
+            warnings.warn('end_time is greater than the MIDI object\'s end time; automatically adjusting to it.')
+            end_time = midi_end_time
+
+        # Enforce that end_time is strictly higher than start_time
+        if start_time >= end_time:
+            warnings.warn('end_time must be strictly higher than start_time; automatically enforcing this.')
```
craffel, line 1381: This should also be a ValueError.

craffel (changes requested):
Thanks for this. Can you wrap to 79 characters, since this package follows PEP 8 to the letter?

pretty_midi/pretty_midi.py
b-boechat: Done.

craffel (approved):

craffel:
Thanks!

pretty_midi/pretty_midi.py
b-boechat: Done.

I thought the same but originally issued a warning by analogy to how adjust_times handles it.

PR #249, opened 2024-08-20
Corruption check before loading tempo changes

TimFelixBeyer:
On some corrupted files, PrettyMIDI gets stuck while calling `_load_tempo_changes`.
This can be avoided by moving the call behind the corruption check.


craffel (approved):

craffel:
Thanks!

PR #252, opened 2024-11-13
Python 3.12 compatibility: remove pkg_resources dependency

roszcz:
`pretty_midi` doesn't work with python 3.12 because of `pkg_resources` deprecation: https://setuptools.pypa.io/en/latest/pkg_resources.html, `import pretty_midi` fails with:

```
ModuleNotFoundError: No module named 'pkg_resources'
```

Replacing `pkg_resources` with `importlib.resources` is not possible (discussed in https://github.com/craffel/pretty-midi/pull/245), because `resources.files` changed in 3.12: https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files – this PR is using `importlib_resources` backport instead: https://pypi.org/project/importlib-resources/.

Code I used to test:

```python
import pretty_midi

midi = pretty_midi.PrettyMIDI("example.mid")
data = midi.fluidsynth()
```

Right now this returns exactly the same array for pythons 3.12 and 3.10.

### DIFF
--- a/pretty_midi/fluidsynth.py
+++ b/pretty_midi/fluidsynth.py
@@ -49,7 +49,7 @@ def get_fluidsynth_instance(synthesizer=None, sfid=0, fs=None):
         raise ImportError("fluidsynth() was called but pyfluidsynth is not installed.")
 
     if synthesizer is None:
-        synthesizer = str(importlib_resources.files(__name__) / DEFAULT_SF2)
+        synthesizer = os.path.join(str(importlib_resources.files(__name__)), DEFAULT_SF2)
 
     # Create a fluidsynth instance if one wasn't provided
     if isinstance(synthesizer, str):
--- a/pretty_midi/fluidsynth.py
+++ b/pretty_midi/fluidsynth.py
@@ -3,7 +3,7 @@
 """
 
 import os
-import pkg_resources
+import importlib_resources
 
 try:
     import fluidsynth
@@ -49,7 +49,7 @@ def get_fluidsynth_instance(synthesizer=None, sfid=0, fs=None):
         raise ImportError("fluidsynth() was called but pyfluidsynth is not installed.")
 
     if synthesizer is None:
-        synthesizer = pkg_resources.resource_filename(__name__, DEFAULT_SF2)
+        synthesizer = str(importlib_resources.files(__name__) / DEFAULT_SF2)
 
     # Create a fluidsynth instance if one wasn't provided
     if isinstance(synthesizer, str):
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -16,9 +16,6 @@ import six
 import pathlib
 from heapq import merge
 
-import os
-import pkg_resources
-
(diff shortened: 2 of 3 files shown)


pretty_midi/fluidsynth.py
```diff
@@ -49,7 +49,7 @@ def get_fluidsynth_instance(synthesizer=None, sfid=0, fs=None):
         raise ImportError("fluidsynth() was called but pyfluidsynth is not installed.")
 
     if synthesizer is None:
-        synthesizer = pkg_resources.resource_filename(__name__, DEFAULT_SF2)
+        synthesizer = str(importlib_resources.files(__name__) / DEFAULT_SF2)
```
craffel, line 52: The "/" here is exploiting the fact that importlib_resources.files(__name__) returns a pathlib.Path right? I *think* we are otherwise compatible with Python versions before 3.4 so perhaps just use 
`os.path.join(str(importlib_resources.files(__name__)), DEFAULT_SF2)`
instead?

roszcz: I think it's not going to work with 3.4 any way because there's f-string usage in `pretty_midi.fluidsynth`:

https://github.com/craffel/pretty-midi/blob/280352a75c53901eb9b0e4f4dffa416ad041fa33/pretty_midi/fluidsynth.py#L67

And it will throw on import:

```
  File "/app/pretty_midi/fluidsynth.py", line 68                                                                                                                                               f"synthesizer sample rate of {synth_fs} doesn't match provided fs of {fs}")
                                                                             ^
SyntaxError: invalid syntax
```

Just removing the f-string is not going to be enough to work with 3.4, I get other errors:

```
      1 import pretty_midi
      2
----> 3 midi = pretty_midi.PrettyMIDI("example.mid")
      4 data = midi.fluidsynth()

/app/pretty_midi/pretty_midi.py in __init__(self, midi_file, resolution, initial_tempo, charset, mido_object)
    114
    115             # Populate the list of key and time signature changes
--> 116             self._load_metadata(midi_data)
    117
    118             # Check that there are tempo, key and time change events

/app/pretty_midi/pretty_midi.py in _load_metadata(self, midi_data)
    239
    240         # We merge the already sorted lists for every track, based on time
--> 241         self.lyrics = list(merge(*tracks_with_lyrics, key=lambda x: x.time))
    242         self.text_events = list(merge(*tracks_with_text_events, key=lambda x: x.time))
    243

TypeError: merge() got an unexpected keyword argument 'key'
```

roszcz: I made this change anyway, as it seems easier to read without knowing how `pathlib.Path` works.

craffel: Ah, I had forgotten that f-strings didn't exist since the dawn of time. Thanks for making the change anyways.

craffel:
Thanks for this! Merging.

keunwoochoi:
nice - @craffel, do you plan to release a newer version to pypi? 

long live pretty_midi!

craffel:
I can if people bug me enough!

PR #254, opened 2025-02-28
fix: Correct Instrument Name Assignment Using MIDI Program Numbers

luigischmitt:
**Fix Instrument Name Assignment in pretty_midi**

- **Description**:

This Pull Request addresses an issue in the pretty_midi library where instrument names were incorrectly assigned using track names instead of the MIDI program number. The previous implementation led to mislabeling instruments when extracting them from MIDI files.

- **Changes Made**:

Adjusted the _load_instruments method to use INSTRUMENT_MAP for correctly naming instruments based on their program number.
Ensured that the __get_instrument function prioritizes program change events for instrument identification rather than relying on track names.

Testing:

The changes were tested with several MIDI files to confirm that instruments are now accurately named according to their MIDI program numbers.
Comparison with results obtained using mido confirmed the correctness of these changes.

- **Impact**:

This fix allows users to correctly retrieve instrument information from MIDI files, improving the usability and accuracy of the pretty_midi library.

image
image
image
image

Previously, the code was incorrectly assigning the instrument name as “Bohemian GM”, which doesn't make sense because it’s a track title, not the name of the instruments themselves. This led to confusion as the extracted instrument names did not accurately reflect the MIDI content.

With the recent changes, the instrument names are now correctly derived from the MIDI program numbers using the INSTRUMENT_MAP. This ensures that the extracted instrument information is accurate and aligns with the MIDI data.

Additionally, a potential enhancement could be to introduce a new parameter in the function. This would allow users to choose between naming instruments based on the program number or retaining the existing method of using the instrument's actual name as the default.

**Please review this PR and provide feedback. Thank you for your support!**

### DIFF
--- a/pretty_midi/pretty_midi.py
+++ b/pretty_midi/pretty_midi.py
@@ -16,6 +16,7 @@ import six
 import pathlib
 from heapq import merge
 
+from .constants import INSTRUMENT_MAP
 from .instrument import Instrument
 from .containers import (KeySignature, TimeSignature, Lyric, Note,
                          PitchBend, ControlChange, Text)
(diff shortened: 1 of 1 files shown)


craffel:
Hi there, I don't see any reason to do this - if a user wants the instrument name corresponding to an instrument's program, they can just do `pretty_midi.constants.INSTRUMENT_MAP[instrument.program]`. In other words, there is already a one-to-one-correspondence between a member of Instrument and the instrument name, and that mapping is already provided by INSTRUMENT_MAP. If you don't store the track name as a member of Instrument (as is currently done), then you lose that information. In many MIDI files, the track name contains finer-grained information about the instrument than the program number itself. If you wanted to add a get_instrument_name convenience function to Instrument that returned `constants.INSTRUMENT_MAP[self.program]` I'd support that, but there's not a lot of meaningful functionality there.

